PR #7, opened 2025-08-29
New SUP - single client multiple devices initial proposal

julienduquesnay-se:
New SUP proposal for enabling a single client to represent multiple devices.

### DIFF
--- a/proposals/single-client-for-multiple-devices.md
+++ b/proposals/single-client-for-multiple-devices.md
@@ -1,0 +1,42 @@
+# Specification Update Proposal
+
+## Owner
+
+Merrill Harriman, Julien Duquesnay
+
+## Summary
+
+We propose to enhance the Workload Fleet Manager to device interface to allow a single client - a gateway service for example - to provide device capabilities, retrieve desired states, etc. on behalf of one or multiple target devices. 
+
+## Reason for proposal
+
+These enhancements will allow devices that can host Margo applications but do not implement the Margo interface to be managed by a Margo compliant WFM, and thus decrease the barrier to entry for the device vendors. 
+
+They could also allow managing devices that required a higher level of network isolation - no direct connection with the outside world for example.
+
+These enhancements may also prepare the way for supporting constrained devices.
+
+## Requirements alignment acknowledgement
+
+This SUP is aligned with the following technical features:
+
+- https://github.com/margo/specification/issues/99
+- https://github.com/margo/specification/issues/101
+
+## Technical proposal
+
+> The SUP's technical details. There must be enough technical details that someone can take the information in this section and implement it on their own.
+> 
+> Complete as part of Phase 3: SUP Technical Development
+
+## Alternatives considered (optional)
+
+> List any alternative solutions considered while working on the SUP and the reason for not choosing them. If the SUP owner knows that there is a risk of a competing SUP, this section can be used to make their case ahead of any potential votes on why their solution is better.
+> 
+> Complete as part of Phase 3: SUP Technical Development
+
+## Rejection reason
+
+> If a SUP is rejected, indicate the reason why it was rejected.
+> 
+> Complete if SUP is rejected at Phase 2: Proposal Creation or Phase 4: Final Decision 


phil-abb:
@margo/proposal-committee / @margo/technical-wg - A new SUP has been proposed for a [Gate 3 check](https://github.com/margo/Margo-collaboration/blob/main/specification_update_processes/process-overview.md#decision-gate-3)

proposals/single-client-for-multiple-devices.md
```diff
@@ -0,0 +1,42 @@
+# Specification Update Proposal
+
+## Owner
+
+Merrill Harriman, Julien Duquesnay
+
+## Summary
+
+We propose to enhance the Workload Fleet Manager to device interface to allow a single client - a gateway service for example - to provide device capabilities, retrieve desired states, etc. on behalf of one or multiple target devices.
```
phil-abb, line 9: Is the device's OTEL collector included in the 'etc."?

proposals/single-client-for-multiple-devices.md
```diff
@@ -0,0 +1,42 @@
...
+
+Merrill Harriman, Julien Duquesnay
+
+## Summary
+
+We propose to enhance the Workload Fleet Manager to device interface to allow a single client - a gateway service for example - to provide device capabilities, retrieve desired states, etc. on behalf of one or multiple target devices. 
+
+## Reason for proposal
+
+These enhancements will allow devices that can host Margo applications but do not implement the Margo interface to be managed by a Margo compliant WFM, and thus decrease the barrier to entry for the device vendors.
```
phil-abb, line 13: It's unclear from reading this if you will be proposing a new Margo-defined way to communicate between the leaf devices and the "gateway" device, or if you are thinking there will be proprietary communication between the leaf devices and "gateway" device. Would you be able to clarify this in the text?

proposals/single-client-for-multiple-devices.md
```diff
@@ -0,0 +1,42 @@
...
+
+These enhancements will allow devices that can host Margo applications but do not implement the Margo interface to be managed by a Margo compliant WFM, and thus decrease the barrier to entry for the device vendors. 
+
+They could also allow managing devices that required a higher level of network isolation - no direct connection with the outside world for example.
+
+These enhancements may also prepare the way for supporting constrained devices.
+
+## Requirements alignment acknowledgement
+
+This SUP is aligned with the following technical features:
```
phil-abb, line 21: This seems like a whole new feature to me, but I'll defer to the discussion you had with Armand about this.

proposals/single-client-for-multiple-devices.md
julienduquesnay-se: No the device's OTEL collector is not included in the "etc.". This SUP focuses on the Fleet Manager API used by the device. If services required by Margo, like the OTEL collector, are not implemented by the device then it would make sense to have them included in the gateway service. This was not the intent of this SUP to address this. We can discuss if you think we should expand the scope of this SUP or maybe create another SUP to address the functionalities that a gateway service must provide.

proposals/single-client-for-multiple-devices.md
julienduquesnay-se: I have added clarification that this SUP is not targeting PR1.

proposals/single-client-for-multiple-devices.md
phil-abb: I don't see these updates in the text.

phil-abb: Never mind, I see the comment now :)

proposals/single-client-for-multiple-devices.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Owner
+
+Merrill Harriman, Julien Duquesnay
+
+## Summary
+
+We propose to enhance the Workload Fleet Manager to device interface to allow a single client - a gateway service for example - to provide device capabilities, retrieve desired states, etc. on behalf of one or multiple target devices. 
+
+The idea is to augment the Workload Fleet Manager interface to enable the connection of this "single client", while the communication between the "single client" and the target devices stays out of the scope of Margo (it could be a proprietary communication).
```
Haishi2016, line 11: I think we just call it "gateway" or "field gateway" without saying "single client". Ideally, the system architecture should support multiple gateways at the same time, so it's not really a "single client" per se.

julienduquesnay-se: I did not mean to imply that this "single client" would be the only client connecting to the WFM, just that one client would connect multiple devices. There could several of these "single clients". Maybe the simplest way to fix this is to drop the "single" and only use "client".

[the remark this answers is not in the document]
ajcraig: > This SUP was more targeted at the second bullet - visible device(s) behind device. But I have no problem including both options if they are both of interest.

Quick clarifying question here, do the visible devices behind the single client fill a Margo device role(standalone device/standalone cluster) or are these custom devices with proprietary runtimes?

- If Margo roles are behind the single client, we could decouple the current 1to1 notion of client and device. Just an idea, we could upgrade the device capabilities and desired state information with unique device ids, seperate from the client id.

julienduquesnay-se:
> I like both the transparent and opaque models, and think both should be on the roadmap. We need to be careful about prioritisation and engage with PMG; while I like both models, I am not sure that either of them should be in the first release. I would personally prioritise opaque over transparent in a second release, but would probably switch that priority if transparent was substantially less work. Paul

@pauldbrooks, I can see how the two options would lead to different products, but in terms of API needs and data exchange with the Workload Fleet Manager I don't think they are that different. I would try to address them together and if we see a strong divergence then split them into two SUPs.

proposals/res/single-client-for-multiple-devices/types-of-gateway.drawio.svg
stormc, line 1: In the "Opaque Gateway" case, could the Device C also have Capabilities, i.e., Capabilities Device C = A + B + C or is that ruled out such that it only represents Device A and Device B towards the WFM?

proposals/single-client-for-multiple-devices.md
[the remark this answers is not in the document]
stormc: I guess the more common case is what you call the "Opaque Gateway": Consider, for example, an Edge box to which you have sensors attached, e.g. via (proprietary) connectivity, and these sensors cannot represent themselves as Margo capable devices to the WFM. In this case, the Edge box exposes the capabilities of itself and all attached sensors to the WFM as one Margo capable device. Anyway, let's see how different the two variants end up and PM decide which to tackle first.

eriknordmark:
One thing we need to consider for PR1 is whether our concepts and definitions (such as device-id vs client-id) are sufficient to allow us to add this down the road. For instance the API SUP from Armand is looking at using client-id instead of device-id. That might make sense for the opaque gateway (which if I squint has similarities with a multi-node kubernetes cluster in that it presents the sum of the device capabilities.)

Not clear to me what the high-level differences between the pass-through and the transparent - one question is whether the gateway would maintain the identifities (such as X.509 certificates) for the devices it is handling or if it will only use its own identity when communicating to the WFM.

But it would make sense to check whether folks thing Armand's SUP doesn't preclude or make it very odd to add these gateways down the road.

Haishi2016 (approved):

PR #16, opened 2025-09-30
Desired state from oci

Silvanoc:


Silvanoc:
FYI: @phil-abb

eriknordmark (dismissed):
Thanks for putting together all these details.

This comment is pretty long and some of the comments reference particular sections and text in the document. I hope that putting it as one comment instead of line-by-line comments is useful and helps to get an overview.

I'm open to discussion and clarifications around this. 

S1: This seems like an outline of possible proposals, but not a single concrete proposal which can be implemented to get interoperability. I wouldn't know where to start implementing it.

Examples:
- "Signal: Update desired" mentions a push approach but provides no useful details of what that is nor how to accomplish interoperability. Without an actual interoperable specification I don't think this belongs in a standard specification.

- The possibility of using TUF is listed, but is that part of the proposal or not? FWIW there is no need for TUF's cryptographic hashes if the artifacts (helm charts etc) end up bring retrieved using OCI, since the OCI image format already includes such hashes of manifests and layers.

 - The "Management Client requests" section says that the URLs include a digest (both in the table with <prefix>/manifests/<digest> and in the text "This URL includes the digest that can be used to validate the integrity of the obtained manifest"). How does the WFMC know which digest to in the URL? This digest will change each time the desired state changes. Is the proposal assuming that something like SSE will be used to specify this digest?
 
An alternative would be to specify how SSE and/or MQTT would be used to do the signaling, including how they would handle intermittent connectivity (SSE depends on a persistent http connection from the client), and the Purdue network model (would MQTT require relays at each boundary in the Purdue model?)

If TUF is part of the proposal then please provide all the details so we can get interoperable implementations of the use of TUF for this purpose.

S2: Security is declared to be out of scope, which is at least honest.
And the only fully specified mechanism is the periodic polling where the proposal notes ""Latest" tag provided by the REST-API returns the digest which unambiguously identifies the desired state document to be pulled. No defense against the above mentioned attacks is provided."

S3: The proposal contains some rollback protection using a version number in the OCI metadata. It says "That way the device only needs to keep track of the last known desired state version and compare it with that of any provided trusted desired state." But the specification does not include a way for the WFMC to establish what is trusted desired state since the integrity protection of the OCI artifact carrying the desired state is out of scope.

I find that total absence of security to not meet the security needs for factory automation systems, in particular since there are standard approaches which do provide integrity protection and rollback protection - the cryptographic hashes used in the OCI specifications for the artifacts. All that requires is a RESTful API for the WFMC to retrieve the those hashes from the WFM. Simpler and secure. And it seems like that is the direction in which the "SSE" text in the proposal is pointing as far as I can tell.

---

U0: This might be a terminology issue. The proposal says that "using an OCI registry is not mandatory." My understanding is that the OCI distribution spec defines the API between an OCI client and an OCI registry, hence by definition there is an OCI registry in the picture. Is the intention to say that you don't have to use an existing OCI registry code base?

U1: I'm assuming this OCI registry falls in the blue half circles which connect the WFMC to the WFM, and is separate from the white half circle connecting the Component Registry to the OCI container runtime.
If that is correct I think it needs to be made more explicit in the proposal.

U2: But what isn't clear is whether this means that this desired state OCI OCI registry must be co-located with the WFM.
This is important because several of the Margo participants have pointed out that in many deployments they want to have a local OCI registry for the artifacts (containers, helm charts, etc)
If such a local OCI registry can be used for desired state, then i think Margo needs to standardize the method and API by which a WFM can propagate the desired state to a local OCI registry. But if the desired state OCI registry must be co-located with the WFM then this becomes a local matter inside the "WFM service" and no standardization is needed. (There might still be significant complexity in building this though.)

---

I'm having a hard time seeing the benefits of this approach.
Depending on the details it might merely add implementation complexity in a "WFM service" or it might also require Margo to standardize a new API for a WFM to be able to interoperate with some local desired state OCI registry.

Note that I see OCI registries as being very useful for artifacts, and it might very well make sense for us to define an OCI ArtifactType for Margo application manifests. Artifacts have the property that they can, and typically are, used by a multitude of devices. But the desired state API is all about a point-to-point relationship between the WFM and a WFMC with no sharing. Thus it feels like trying to fit a square peg in a round hole.
(Note that using OCI for the Component Registry white half-circle and being able to store Margo Manifests, Compose files, and Helm charts in that registry makes a ton of sense. But here we are talking about the desired state for a single device/WFMC.)

And a NIT, but nothing in the OCI distribution specifications claims it is a RESTful API. https://github.com/opencontainers/distribution-spec/ says "The OCI Distribution Spec project defines an API protocol to facilitate and standardize the distribution of content." Nothing about the State Transfer in REST.
This observation relates to my concern about the fitness for purpose to use OCI for the desired state.

It seems like if the goal is something like the SSE notification service, we would need to specify 1) the payload it will carry and 2) what happens when the WFMC reconnects after a network outage. It turns out that this is actually all we need to specify to have a complete RESTful API for fetching the desired state from the WFM. So the use of OCI doesn't seem to add anything but complexity. What am I missing?

submitted/sup_desired_state_from_oci.md
```diff
@@ -0,0 +1,351 @@
...
+There are multiple open-source software implementations of both registries and client libraries that are battle-tested and widely used across the industry, eliminating the need to develop and maintain custom solutions.
+
+### Benefits of using OCI for desired state
+
+* OCI blobs are being used increasingly for distributing content. It has been used for a long time for distributing container images and Helm charts, and is starting to be used more and more to distribute other types of content. Margo is already planning on using OCI blobs to distribute Helm charts and Docker images.
+* There are many OCI compatible servers, clients, tools, and SDKs that can be used instead of needing to implement anything Margo specific.
+* There are existing patterns for signing OCI blobs that can be applied
+* We are enabling the use of existing standards and implementations instead of inventing something Margo-specific that would require a custom implementation.
+* Placing desired state content in an OCI blob is more flexible/future-proof. If we have more that needs to be provided in the future than the desired state yaml files, the same approach will work without needing to change APIs.
```
eriknordmark: I don't think this matches how OCI is used elsewhere. OCI is used for this which can be published and then is possible to be referenced and used by multiple parties. Container images, Helm charts, Margo manifests all fall in that category. The desired state for a single device does not.

submitted/sup_desired_state_from_oci.md
```diff
@@ -0,0 +1,351 @@
...
+It is important to remember that the decision to use a REST API instead of Git (which the specification currently describes) to distribute the desired state was **voted on by the Margo TWG members**.
+It is therefore considered out of discussion in this proposal.
+
+The following two approaches can be used to achieve the desired state of one or more applications on an edge device:
+
+1. The edge device fetches (pulls) the new desired state documents.
+   This requires a specification of the format and protocol to distribute those documents, which includes the REST API and the "packaging".
+   Check the section [Fetch desired state: REST API](#fetch-desired-state-rest-api) for more details on this.
+1. The edge device is signaled (notification from the backend, internal schedule for polling,...) to check the availability of new desired state document(s).
+   Check the section [Signal: Update desired state](#signal-update-desired-state) for more details on this.
```
eriknordmark: This reads like speculation and not part of a specification. Please either complete the specification of SSE including the payload format use, or drop it from the proposal.

submitted/sup_desired_state_from_oci.md
```diff
@@ -0,0 +1,351 @@
...
+This specification only proposes implementing the pull approach, but does not prevent a notification approach from being proposed in the future as an enhancement.
+
+This proposal is very similar to the previous Git approach and represents almost a full-blown GitOps approach using OCI blobs instead of Git.
+
+### REST-API specification
+
+The hereby proposed "well-established REST API" is that of the [OCI Distribution v1.1 specification](https://github.com/opencontainers/distribution-spec/blob/v1.1.0/spec.md).
+
+As mentioned before, there are multiple implementations of this API (both servers and clients), many of them open source software (OSS).
+This has the benefit of allowing available OCI registries and client libraries to be used (e.g., ORAS, FluxCD, ArgoCD). Since we would only be adopting the APIs it gives WFM vendors the freedom to either use existing OCI registry solutions or implement their own custom backend solution without requiring the use of an OCI registry.
```
eriknordmark: As I said elsewhere, this use of "OCI registry" is inconsistent with the distrbution specification. Anything which implements the "server" side of that specification is defined to be an OCI registry.

submitted/sup_desired_state_from_oci.md
```diff
@@ -0,0 +1,351 @@
...
+See below for more information
+
+### Management client requests
+
+It MUST be possible for the device to pull the manifest and blob using the behavior defined in the [Open Container Initiative Distribution Specification](https://github.com/opencontainers/distribution-spec/blob/main/spec.md) for the following endpoints.
+
+|  |  |  |
+|---|---|---|
+| `GET`/`HEAD` | `<prefix>/manifests/<digest>`| Used to pull the manifest describing the device's desired state|
+| `GET`/`HEAD` | `<prefix>/blobs/<digests>`| Used to pull the blob containing the device's desired state|
```
eriknordmark: The proposed specification update does not say where this digest comes from. So the specification looks self-inconsistent.

eriknordmark (commented):
In addition to detailed line by line comments¸ one thing missing from this specification are the ArtifactType strings used for the Margo desired state artifact. Without this specified you can't have interoperable implementations.

The base OCI artifactType is vnd.oci.image.manifest.v1+json (and I can't find one for helm charts right now).

submitted/sup_desired_state_from_oci.md
phil-abb: This is why one of the benefits this SUP lists is that tools like Flux and Argo can be used because they support this type of workflow.

As I showed in my video linked in the SUP examples, you can use existing Flux or Argo operators on the device (if you want to) to apply the desired state without having to implement a complete solution on your own.

submitted/sup_desired_state_from_oci.md
```diff
@@ -181,7 +181,7 @@ It MUST be possible for the device to pull the manifest and blob using the behav
 
 |  |  |  |
 |---|---|---|
-| `GET`/`HEAD` | `<prefix>/manifests/<digest>`| Used to pull the manifest describing the device's desired state|
+| `GET`/`HEAD` | `<prefix>/manifests/<reference>`| Used to pull the manifest describing the device's desired state|
 | `GET`/`HEAD` | `<prefix>/blobs/<digests>`| Used to pull the blob containing the device's desired state|
```
eriknordmark: Two lines had <digests> in the URL (and my comment was against both).

submitted/sup_desired_state_from_oci.md
```diff
@@ -0,0 +1,351 @@
...
+| `GET`/`HEAD` | `<prefix>/manifests/<digest>`| Used to pull the manifest describing the device's desired state|
+| `GET`/`HEAD` | `<prefix>/blobs/<digests>`| Used to pull the blob containing the device's desired state|
+
+If an OCI registry is used without an API gateway, then `<prefix>` must be `/v2/<name>`. Where `<name>` is the name assigned to the repository in the OCI registry.
+
+How `<prefix>` is provided to the device is outside the scope of this proposal, but the expectation is that the Workload Fleet Manager provides it during onboarding.
+
+#### Manifest request
+
+The Workload Fleet Manager MUST provide the URL that the device must call to obtain the manifest (this is expected to be done during onboarding).
```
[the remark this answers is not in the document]
eriknordmark: As long as it is clear that they are architecturally separate registries (assuming the application catatog is also using OCI) it is clear. The fact that one can configure them to point at the same registry is a detail from an architectural perspective.

submitted/sup_desired_state_from_oci.md
```diff
@@ -0,0 +1,351 @@
...
+
+### Fetch desired state: REST API
+
+In the case of polling, no information about the latest desired state version is available on the device. As a first step, the "latest" tag will be used to identify the desired state blob needing to be pulled down by the device.
+
+> Note: See the [Security Considerations (Out of scope)](#security-considerations-out-of-scope) section below for more information about how this process can be made more secure in future enhancements. For PR1 it [has been agreed](https://github.com/margo/specification/issues/44) by the PM Group that we do not need to address all security concerns up front, which isn't to say we should ignore them. The Security Considerations section highlights how this process can be made more secure with future enhancements, so we can focus on first getting a replacement for the Git approach that the specification currently defines, which has been rejected by the TWG, within the short deadline of the end of 2025 for PR1.
```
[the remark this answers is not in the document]
eriknordmark: I don't think we should invent any new security **mechanism** for Margo.
But we do need a security **architecture** that fits the threat model and deployment models we want to support (such as intermittent connectivity, local registries etc).

As I commented elsewhere I think the benefits of signed images and artifacts is important for the end-to-end ecosystem, but that the benefit of it is the desired state RESTful API is limited if not zero, given the http signature proposed in #18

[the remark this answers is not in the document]
eriknordmark: > Also, we know OCI artifacts have digests, so if we determine at some point that using tags isn't going to be enough, then we have the option to make modifications to the spec to make the digests available.

I'm concerned that would be close to going back to square 1. Security is very hard to add after the fact.

submitted/sup_desired_state_from_oci.md
eriknordmark: That Argo CD demo video doesn't give any hints what it would look like when there are 10 (or 10,000) clusters which use the same assets. So it doesn't help me understanding their design center.

But more importantly, maybe I missed something major.

Is this SUP implying that a device would be running Argo CD and/or Flux, thus that they would be requires components of a compliant WFMA? Or are the WFMA only required to have some OCI compliant distribution spec client (where ORAS is an example implementation).

I've been assuming that any Argo CD or Flux (just like our previous discussion about git and GitOps) can useful tools to have in the backend of an implementation of a WFM, but without any impact on the footprint and complexity of what's running on the Margo devices.

phil-abb: > Is this SUP implying that a device would be running Argo CD and/or Flux, thus that they would be required components of a compliant WFMA? Or are the WFMA only required to have some OCI-compliant distribution spec client (where ORAS is an example implementation).

The SUP indicates by using the OCI approach, there are existing tools that **could** be used if the device vendor wants to use them, since they support pulling and applying manifests packaged as OCI blobs in OCI registries. Same on the WFM side, there are existing off-the-shelf OCI registries that **could** be used if the WFM vendor wants, but they can still implement their own custom solution.

> There are many OCI compatible servers, clients, tools, and SDKs that **can** be used instead of needing to implement anything Margo specific.

> As mentioned before, there are multiple implementations of this API (both servers and clients), many of them open source software (OSS). This has the benefit of **allowing** available OCI registries and client libraries to be used (e.g., ORAS, FluxCD, ArgoCD). Since we would only be adopting the APIs it gives WFM vendors the freedom to **either use existing** OCI registry solutions or **implement their own custom** backend solution...

submitted/sup_desired_state_from_oci.md
eriknordmark: FWIW a specification can't require "at least" unless you specify sufficient detail for the optional choices for interoperable implementations. It can require "only" for what you've specified in detail.

submitted/sup_desired_state_from_oci.md
phil-abb: Resolving the conversation since the SUP was not approved

submitted/sup_desired_state_from_oci.md
phil-abb: Resolving the conversation since the SUP was not approved

submitted/sup_desired_state_from_oci.md
```diff
@@ -0,0 +1,351 @@
...
+The following is in scope for this proposal:
+
+1. Packaging and distributing the desired state documents
+1. Using the APIs defined in the [REST-API Specification](#rest-api-specification) section for pulling the desired state to the device
+1. Provide as much security as possible with the current scope and provide extension points for advanced security.
+
+The following is not in scope for this proposal:
+
+1. Defining mechanisms for edge device authentication and authorization to fetch the desired state documents.
+1. Ensuring that only specific versions of valid (authenticity and integrity) desired state documents can be installed to avoid installation of vulnerable versions
```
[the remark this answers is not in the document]
phil-abb: Resolving the conversation since the SUP was not approved

submitted/sup_desired_state_from_oci.md
phil-abb: Resolving the conversation since the SUP was not approved

submitted/sup_desired_state_from_oci.md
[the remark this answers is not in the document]
phil-abb: Resolving the conversation since the SUP was not approved

submitted/sup_desired_state_from_oci.md
```diff
@@ -0,0 +1,351 @@
...
+
+### Fetch desired state: REST API
+
+In the case of polling, no information about the latest desired state version is available on the device. As a first step, the "latest" tag will be used to identify the desired state blob needing to be pulled down by the device.
+
+> Note: See the [Security Considerations (Out of scope)](#security-considerations-out-of-scope) section below for more information about how this process can be made more secure in future enhancements. For PR1 it [has been agreed](https://github.com/margo/specification/issues/44) by the PM Group that we do not need to address all security concerns up front, which isn't to say we should ignore them. The Security Considerations section highlights how this process can be made more secure with future enhancements, so we can focus on first getting a replacement for the Git approach that the specification currently defines, which has been rejected by the TWG, within the short deadline of the end of 2025 for PR1.
+
+#### Rollback attacks
```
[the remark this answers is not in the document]
phil-abb: Resolving the conversation since the SUP was not approved

submitted/sup_desired_state_from_oci.md
```diff
@@ -0,0 +1,351 @@
...
+
+1. Packaging and distributing the desired state documents
+1. Using the APIs defined in the [REST-API Specification](#rest-api-specification) section for pulling the desired state to the device
+1. Provide as much security as possible with the current scope and provide extension points for advanced security.
+
+The following is not in scope for this proposal:
+
+1. Defining mechanisms for edge device authentication and authorization to fetch the desired state documents.
+1. Ensuring that only specific versions of valid (authenticity and integrity) desired state documents can be installed to avoid installation of vulnerable versions
+1. Securing the authenticity and integrity of the individual desired state documents (via signatures)
```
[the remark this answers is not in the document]
phil-abb: Resolving the conversation since the SUP was not approved

phil-abb (approved):

PR #22, opened 2025-10-06
Issue 89 - Application Registry as OCI Registry - Tech Proposal

arne-broering:
This is a draft of the technical proposal for this SUP.


arne-broering:
> @arne-broering - For this PR, the proposal markdown should be moved from the /Proposal to the /Submitted folder

done.

PR #38, opened 2026-02-05
Margo Identity and Authorization Framework (MIAF)

matlec:
First draft of the "Margo Identity and Authorization Framework" (MIAF) SUP for community feedback.

Background: [Thoughts on Identity and Interoperability in Margo - From PR1 to GA](https://docs.google.com/document/d/1DGvXXlSty79v_PFTRvt8Bw2pspfqP13b3PBaSIaL0IQ/edit?usp=sharing)


proposals/margo-identity-and-authorization-framework.md
```diff
@@ -0,0 +1,2172 @@
...
+- The same LDI **MUST NOT** be active for multiple PDIs concurrently.
+
+**Relationship to MIAF:**
+The X.509 SVID Profile defines *how* any Margo Identity is expressed as an X.509 certificate; this device profile defines *additional, device-specific constraints* on that expression.
+
+#### Profile-specific Enrollment and Identity Issuance
+
+Device enrollment uses the generic API defined in [Section 5](#enrollment-and-identity-issuance-endpoint) with the following constraints:
+
+- The only permitted `svid_profile_uri` for devices is `https://margo.org/profiles/spiffe/x509-svid/v1`. Attempts to enroll a device with `jwt-svid` **MUST** be rejected with `422` (`unsupported-svid-profile`).
```
phil-abb: I might just be getting confused, but this seems like a contradiction to [what is below](https://github.com/margo/specification-enhancements/blob/feat/miaf-sup/proposals/margo-identity-and-authorization-framework.md#bootstrap-method-flows). Below indicates that either X. 509 or JWT can be used for enrollment.

proposals/margo-identity-and-authorization-framework.md
```diff
@@ -0,0 +1,2172 @@
...
+- Device enrollment **MUST** use one of the device bootstrap methods in [Appendix A](#appendix-a-bootstrap-methods-normative) (e.g., **FDO**, **IEEE 802.1AR DevID**, **factory certificate** via mTLS or JWT assertion).
```
phil-abb: Do we need to indicate that the server and device MUST support at least the Factory Certificate Method? If we don't indicate at least one that the server and device MUST support, then it seems like it's going to be a problem for interoperability. They could both support others, but we need at least a common one.

matlec:
> @matlec I'm not too familiar with SPIFFE yet. How much of this will work out of the box with an existing SPIFFE implementation, and how much of this is a custom Margo implementation?

The SUP reuses SPIFFE for identity naming (SPIFFE IDs), credential formats (X.509-SVID, JWT-SVID), and trust distribution (Trust Bundles). Existing SPIFFE libraries and tooling can be used for SVID validation, Trust Bundle handling, and SPIFFE ID processing. Everything else - device bootstrap methods, the discovery document, enrollment / renewal / revocation APIs, the LDI / PDI / ESI binding model, and device replacement / lifecycle rules - is Margo-specific and would need to be implemented on top. The revised draft now includes a "Relationship to SPIFFE" section with a classification table that makes this boundary explicit.

nilanjan-samajdar (commented):
High level comments for now on overall strategy for adopting this SUP and demarcation between the specification and implementation, will review from API perspective on the latest version of the SUP.

proposals/margo-identity-and-authorization-framework.md
```diff
@@ -0,0 +1,2227 @@
...
+Examples include the **Device Fleet Manager (DFM)**, **Workload Fleet Manager (WFM)**, their clients, and infrastructure services such as registries or observability collectors.
+This SUP concerns identities used by these *non-human* components.
+
+#### Workload Fleet Management (WFM) Client (relationship to MIAF) <!-- omit from toc -->
+
+A **Margo client component** as defined in the [Technical Lexicon](https://specification.margo.org/personas-and-definitions/technical-lexicon/). While a WFM Client runs on an Edge Compute Device, its identity represents the deployed **client instance**, not the device itself. The **Logical Device Identity** defined in this SUP provides the stable, hardware-bound identity of the device. A planned **WFM Client Identity Profile** will define how WFM Clients obtain their own distinct Margo Identities, building on the device identity as their authentication foundation. This separation is necessary because device identity and WFM Client identity have different lifecycles, authorization scopes, and cardinalities across topologies (standalone devices, Kubernetes clusters, device gateways).
```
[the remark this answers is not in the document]
matlec: Hey @nilanjan-samajdar! Please review the [Future Work](https://github.com/margo/specification-enhancements/blob/753669de83484d3516eb0505b7b338d586db7151/proposals/margo-identity-and-authorization-framework.md#9-future-work-wfm-client-identity-profile-informative) section that I added to the SUP since you left your comment. Does that resolve your question?

proposals/margo-identity-and-authorization-framework.md
```diff
@@ -0,0 +1,2227 @@
...
+#### Enrollment Subject Identifier (ESI) <!-- omit from toc -->
+
+A deterministic, globally unique identifier derived by the MIS from the presented **Bootstrap Credential** during enrollment.
+It is used to decide whether the presented bootstrap proof corresponds to an existing **Margo Identity** or a new one.
+
+The derivation is **method-specific** and defined by each **Bootstrap Method**.
+*Example (device profile):* from a verified PDI, the ESI may be the certificate fingerprint, or a hash derived from a device certificate contained in an FDO Ownership Voucher.
+ESIs **MUST** be stable and unique within the Trust Domain and **MUST NOT** be reversible to the original credential material.
+
+#### SPIFFE Verifiable Identity Document (SVID) <!-- omit from toc -->
```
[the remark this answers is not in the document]
matlec: I'm not exactly sure I understood your question. The ESI serves multiple purposes in the framework:
1. It enables idempotent enrollment of devices / identities (it makes enrollment safe to retry across network failures without creating duplicate identities).
2. It gives the MIS a single consistent key to look up whether it has seen this device before regardless of *how* the device is authenticated (per bootstrap method).
3. It enables swapping the hardware while preserving the logical identity. A replacement device brings a new PDI (which will produce a different ESI), which can then be bound to an existing LDI.

I think of an ESI as a stable, bootstrap method-independent, privacy-preserving key that lets the MIS answer the question: "is this the same physical device?" without being coupled to any particular bootstrap method / protocol.

proposals/margo-identity-and-authorization-framework.md
```diff
@@ -0,0 +1,2227 @@
...
+For devices, it conveys evidence of the **Physical Device Identity**. Each supported **Bootstrap Method** defines how this credential is formatted and verified.
+
+#### Bootstrap Method <!-- omit from toc -->
+
+A pluggable, normative method by which a Margo component presents its **Bootstrap Credential** to the MIS for enrollment.
+This SUP defines methods for Edge Compute Devices, including **FDO**, **factory certificate** (via mTLS or JWT assertion), and **IEEE 802.1AR DevID**. Future SUPs may introduce methods for other Margo components.
+
+#### Enrollment Subject Identifier (ESI) <!-- omit from toc -->
+
+A deterministic, globally unique identifier derived by the MIS from the presented **Bootstrap Credential** during enrollment.
```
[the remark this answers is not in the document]
matlec: The ESI should remain internal to the MIS as it is an concept of the enrollment layer, and exposing it would couple other components to bootstrap internals. That said, I agree that the use cases you describe are real operational needs. But I'd rather address them via LDI lifecycle events (revocation, re-issuance). The WFM cares that spiffe://.../margo/device/<uuid> was revoked or re-issued, not which ESI was involved.

For now, I'd treat this as a gap that can be addressed by future work (e.g., by defining a lifecycle event/notification mechanism). Or do you think this is critical to include in a first iteration?

proposals/margo-identity-and-authorization-framework.md
```diff
@@ -0,0 +1,2200 @@
...
+   - public keys used to verify JWT SVIDs (if used).
+
+   Trust Bundles are identified by their Trust Domain name and distributed using the SPIFFE **Bundle Map mechanism**, as defined in the [SPIFFE Trust Domain and Bundle specification](https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Trust_Domain_and_Bundle.md).
+   Margo clients and servers use these bundles when validating SVIDs during mutual authentication.
+
+A typical interaction sequence in MIAF (applicable to all Margo components, not just devices) is as follows. This sequence is **informative** and illustrates a typical flow; normative API details are in [Section 5](#5-apis).
+
+1. **Discovery:** A Margo component locates MIS endpoints and Trust Bundle locations via the `.well-known/margo` discovery document defined in this SUP.
+2. **Enrollment:** The component presents a **Bootstrap Credential** (per its Bootstrap Method) to the MIS and receives an **SVID** for its identity.
+3. **Renewal:** Before expiry, the component renews its SVID via an authenticated request (for example, mTLS using the current SVID).
```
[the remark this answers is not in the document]
matlec: See my comment in [this thread](https://github.com/margo/specification-enhancements/pull/38#discussion_r3087394230).

proposals/margo-identity-and-authorization-framework.md
```diff
@@ -0,0 +1,2200 @@
...
+   - maintaining the binding between **Physical Device Identities (PDIs)** and **Logical Device Identities (LDIs)** for devices covered by this SUP.
+
+   MIS exposes a consistent set of APIs - **discovery**, **enrollment**, **renewal**, **revocation**, and **JWT SVID exchange** - that all identity profiles build upon.
+   This API design allows other components (e.g., WFM Clients or telemetry agents) to reuse the same trust foundation in future SUPs.
+
+3. **Margo components using MIAF**
+   
+   Components such as DFMs, WFMs, their clients, and telemetry agents act as:
+
+   - **SVID holders**, presenting their SVIDs during mTLS authentication or when requesting a short-lived **JWT SVID**; and
```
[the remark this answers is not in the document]
matlec: I have updated the SUP since you left your comment to clarify the relationship with SPIFFE and make it more explicit. Please check if this [updated section](https://github.com/margo/specification-enhancements/blob/753669de83484d3516eb0505b7b338d586db7151/proposals/margo-identity-and-authorization-framework.md#relationship-to-spiffe-) is sufficient / addresses your comment.

proposals/margo-identity-and-authorization-framework.md
```diff
@@ -0,0 +1,2200 @@
...
+   - **verifiers**, validating peer SVIDs using the Trust Domain's **Trust Bundle**, then applying **policy-based authorization** based on verified **SPIFFE IDs** and attributes.
+
+4. **Trust Bundles**
+   
+   Each Trust Domain publishes a **Trust Bundle** containing:
+
+   - root and intermediate certificates used to validate X.509 SVID chains; and
+   - public keys used to verify JWT SVIDs (if used).
+
+   Trust Bundles are identified by their Trust Domain name and distributed using the SPIFFE **Bundle Map mechanism**, as defined in the [SPIFFE Trust Domain and Bundle specification](https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Trust_Domain_and_Bundle.md).
```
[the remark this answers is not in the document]
matlec: No, there is a single trust domain for an entire deployment which would include WFM-related identities (once we have established a MIAF-compatible concept for them, see [Future Work](https://github.com/margo/specification-enhancements/blob/753669de83484d3516eb0505b7b338d586db7151/proposals/margo-identity-and-authorization-framework.md#9-future-work-wfm-client-identity-profile-informative))

proposals/margo-identity-and-authorization-framework.md
```diff
@@ -0,0 +1,2200 @@
...
+> The HTTPS connection used to retrieve the Trust Bundle **MUST** be authenticated using an initial trust mechanism as defined in [Initial Trust Bootstrap](#initial-trust-bootstrap).
+> The SPIFFE Bundle Map format supports inclusion of bundles for multiple Trust Domains, which can serve as a basis for cross-domain trust. Full federation lifecycle semantics, as defined by SPIFFE Federation, are out of scope for this SUP.
+
+#### Enrollment and Identity Issuance Endpoint
+
+This endpoint is used by a Margo component (for this SUP: an Edge Compute Device) or an authorized bootstrap intermediary acting on its behalf (for example, an FDO Owner Onboarding Service component of the MIS) to perform **initial enrollment** with the Margo Identity Service (MIS).
+
+During enrollment, the component authenticates using its **Bootstrap Credential** and requests issuance of a new identity, represented by an SVID.
+For Edge Compute Devices, this operation establishes the authoritative binding between the device's **Physical Device Identity** and **Logical Device Identity** within the Trust Domain.
+
```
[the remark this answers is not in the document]
matlec: I don't have a plan for now, but if we agree on [Silvano's SUP](https://github.com/margo/specification-enhancements/pull/48), I will not resist to update any LinkML specs while integrating this SUP into the spec once/if approved :)

proposals/margo-identity-and-authorization-framework.md
```diff
@@ -0,0 +1,2200 @@
...
+| :---- | :--- | :------- | :---------- |
+| `svid_profile_uri` | string | Y | Absolute URI identifying the SVID profile requested. **MUST** match one of the URIs listed in `svid_profiles_supported` from the [discovery document](#discovery-document-endpoint). |
+| `svid_request` | object | Y | Profile-specific payload containing parameters required to issue an SVID. See [Profile-specific `svid_request` formats (request payload)](#profile-specific-svid_request-formats-request-payload) below. |
+| `bootstrapCredential` | object | Y | Credential and associated proof used to authenticate the component during enrollment. See [Bootstrap Methods](#appendix-a-bootstrap-methods-normative) for normative method definitions. |
+| `bootstrapCredential.method` | string | Y | URN uniquely identifying the bootstrap method (e.g., `urn:margo:bootstrap:factory-cert-jwt:v1`). |
+| `bootstrapCredential.proof`  | object | N | Method-specific proof of possession (e.g., a signed JWT assertion or an mTLS client certificate chain). Present only if the bootstrap method requires explicit proof material. |
+
+**Response body schema (`201 Created` or `200 OK`, `application/json`):**
```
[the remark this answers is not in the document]
matlec: See my answer in [this thread](https://github.com/margo/specification-enhancements/pull/38#discussion_r3087434643)

matlec:
Hey @margo/technical-wg!

I made a couple of changes to the SUP over the last days and would love to hear your feedback.

1. I completed the integration of the [FDO bootstrap method](https://github.com/margo/specification-enhancements/blob/feat/miaf-sup/proposals/margo-identity-and-authorization-framework.md#fido-device-onboard-fdo-method) in the framework to support onboarding edge devices based on [FIDO Device Onboard](https://fidoalliance.org/specs/FDO/FIDO-Device-Onboard-PS-v1.1-20220419/FIDO-Device-Onboard-PS-v1.1-20220419.html).
2. I removed the IEEE 802.1AR bootstrap method from the SUP. The 802.1AR standard does only define requirements for the format (and secure storage) of a "birth certificate", not an onboarding protocol. Hence, the existing certificate-based bootstrap methods already support the use of IDevID certificates and standardizing on e.g. IDevIDs can be a policy decision within a Trust Domain. I added an [informative note](https://github.com/margo/specification-enhancements/blob/feat/miaf-sup/proposals/margo-identity-and-authorization-framework.md#using-ieee-8021ar-devids-with-bootstrap-methods-informative) to the SUP that reflects this decision.
3. I added another bootstrap method based on a [single-use enrollment token](https://github.com/margo/specification-enhancements/blob/feat/miaf-sup/proposals/margo-identity-and-authorization-framework.md#enrollment-token-method) to the SUP to address concerns on how to integrate brownfield, low-cost, and constrained devices. @stormc This is related to your request for bootstrapping devices based on a pre-shared key. Please review the enrollment token method I added and provide feedback whether this addresses the concerns you expressed.

Thanks everyone for the feedback you provided so far!

chrisgclayton:
I am personally in alignment with the concept of SPIFFE. The questions I would have on this are around when doing the outside in interrogation to determine identities do we have any additional that we require for our purposes?

chrisgclayton:
If we are going to use the SPIFFE id which is generally used for workloads vs. device ids I would recommend we do not recommend it is the same trust domain (device and workload) to ensure we do not have conflicts. An alternative is to use the same format with different scheme (i.e., instead of SPIFFE://trustdomain/path we use MARGO://trustdomain/path).

matlec:
> I am personally in alignment with the concept of SPIFFE. The questions I would have on this are around when doing the outside in interrogation to determine identities do we have any additional that we require for our purposes?

First of all, thanks for reviewing the SUP! If by

> outside in interrogation to determine identities

you mean the attestation step used to determine what identity a caller should receive, then I think that is the right question. My understanding is that SPIFFE gives us the identity primitives, SVID formats, and trust-domain model, but it does not standardize the local interrogation or attestation mechanism itself. This is implementation-specific and in the case of SPIRE addressed by node and workload attestation. For the SUP, the device profile is intentionally operating one layer above that: it defines additional Margo-specific requirements for remote device bootstrap and lifecycle (including discovery, standardized bootstrap methods, binding validated bootstrap evidence to a persistent device identity, renewal, replacement, and revocation). So, yes, we do have additional requirements for our purposes, but they are mainly around device bootstrap and lifecycle rather than changing the SPIFFE identity model itself.

If we later define workload- or WFM Client-style identities (WFM Client identities rather sooner than later ;)), we should be explicit about whether local attestation remains implementation-specific or whether Margo wants to standardize that layer as well.

I hope this addresses your concern...

matlec:
> If we are going to use the SPIFFE id which is generally used for workloads vs. device ids I would recommend we do not recommend it is the same trust domain (device and workload) to ensure we do not have conflicts. An alternative is to use the same format with different scheme (i.e., instead of SPIFFE://trustdomain/path we use MARGO://trustdomain/path).

I would prefer to keep the `spiffe://` scheme rather than introducing a `margo://` scheme (actually, I introduced and dropped a `margo:` scheme in an earlier version of this SUP). The main value of this approach is that we are reusing the existing SPIFFE standard, its concepts, validation semantics, ... so that we benefit from existing tooling and implementations.
Regarding shared trust domains, I actually think devices and workloads being in the same trust domain can be the right model for a deployment. The trust domain should represent the shared security boundary, while different principal types (devices, workloads, WFM clients, ...) are separated by distinct SPIFFE ID path namespaces (e.g. `/margo/device/...` for devices) and authorization policy. That lets a device use its SVID to authenticate as a device to the MIS or to other workloads, and makes patterns like issuing a WFM Client identity bound to an authenticated device identity straightforward, without requiring separate trust domains or a non-SPIFFE URI scheme. So I wouldn't enforce separation of trust domains per spec, but rather treat that as a deployment decision. But happy to further discuss this and learn more about your concerns and thoughts.

chrisgclayton:
> > I am personally in alignment with the concept of SPIFFE. The questions I would have on this are around when doing the outside in interrogation to determine identities do we have any additional that we require for our purposes?
> 
> First of all, thanks for reviewing the SUP! If by
> 
> > outside in interrogation to determine identities
> 
> you mean the attestation step used to determine what identity a caller should receive, then I think that is the right question. My understanding is that SPIFFE gives us the identity primitives, SVID formats, and trust-domain model, but it does not standardize the local interrogation or attestation mechanism itself. This is implementation-specific and in the case of SPIRE addressed by node and workload attestation. For the SUP, the device profile is intentionally operating one layer above that: it defines additional Margo-specific requirements for remote device bootstrap and lifecycle (including discovery, standardized bootstrap methods, binding validated bootstrap evidence to a persistent device identity, renewal, replacement, and revocation). So, yes, we do have additional requirements for our purposes, but they are mainly around device bootstrap and lifecycle rather than changing the SPIFFE identity model itself.
> 
> If we later define workload- or WFM Client-style identities (WFM Client identities rather sooner than later ;)), we should be explicit about whether local attestation remains implementation-specific or whether Margo wants to standardize that layer as well.
> 
> I hope this addresses your concern...

Yes I mean the attestation step which to your point is addressed in SPIRE. 

I believe in the future we need to get to workload and agree we should be explicit about it.

chrisgclayton:
> > If we are going to use the SPIFFE id which is generally used for workloads vs. device ids I would recommend we do not recommend it is the same trust domain (device and workload) to ensure we do not have conflicts. An alternative is to use the same format with different scheme (i.e., instead of SPIFFE://trustdomain/path we use MARGO://trustdomain/path).
> 
> I would prefer to keep the `spiffe://` scheme rather than introducing a `margo://` scheme (actually, I introduced and dropped a `margo:` scheme in an earlier version of this SUP). The main value of this approach is that we are reusing the existing SPIFFE standard, its concepts, validation semantics, ... so that we benefit from existing tooling and implementations. Regarding shared trust domains, I actually think devices and workloads being in the same trust domain can be the right model for a deployment. The trust domain should represent the shared security boundary, while different principal types (devices, workloads, WFM clients, ...) are separated by distinct SPIFFE ID path namespaces (e.g. `/margo/device/...` for devices) and authorization policy. That lets a device use its SVID to authenticate as a device to the MIS or to other workloads, and makes patterns like issuing a WFM Client identity bound to an authenticated device identity straightforward, without requiring separate trust domains or a non-SPIFFE URI scheme. So I wouldn't enforce separation of trust domains per spec, but rather treat that as a deployment decision. But happy to further discuss this and learn more about your concerns and thoughts.

I am ok with it being the same trust domain as long as the pathing is clearly different. I want to be sure there is never an authorization decision that is workload related and based on the device identity, or vice versa.

proposals/margo-identity-and-authorization-framework.md
```diff
@@ -0,0 +1,2563 @@
...
+   - **verifiers**, validating peer SVIDs using the Trust Domain's **Trust Bundle**, then applying **policy-based authorization** based on verified **SPIFFE IDs** and attributes.
+
+4. **Trust Bundles**
+   
+   Each Trust Domain publishes a **Trust Bundle** containing:
+
+      - X.509 trust-anchor certificates used to validate X.509 SVID chains; and
+      - public keys used to verify JWT SVIDs (if used).
+
+      Intermediate CA certificates **MAY** also be distributed by the deployment, including alongside issued SVID chains, but they are not required to appear in the Trust Bundle when standard path-building succeeds without them.
```
phil-abb: There are two statements about Intermediate CA certificates that are unclear to me. 

This statement:
>"Intermediate CA certificates MAY also be distributed by the deployment, including alongside issued SVID chains, but they are not required to appear in the Trust Bundle when standard path-building succeeds without them.

and another one further down the page:
> For X.509 authorities published in a SPIFFE Bundle Map, the bundle represents the public trust anchors for that Trust Domain. Intermediate CA certificates MAY also be published, but are not required when they are conveyed with issued SVID chains or distributed by other deployment-specific means.

Since both of these statements are "MAY" statements its not clear to me what the impact either doing or not doing this may have. Is the impact of this completely within the black box of the MIS implementation? Or will the decision on how this is handled have an impact on what the device, WFM, DFM, would need to do if certain choices are made?

proposals/margo-identity-and-authorization-framework.md
```diff
@@ -0,0 +1,2563 @@
...
+| **Active** | The device uses its valid SVID to authenticate to Margo components within the Trust Domain. |
+| **Renewal** | Before expiry, the device renews its SVID via an authenticated request (e.g., mTLS with the current SVID). Renewal semantics, including rate-limiting and backoff, are defined in [SVID Renewal Endpoint](#svid-renewal-endpoint). |
+| **Replacement** | When the logical identity must persist across hardware replacement or another operator-authorized rebinding, MIS binds a **new** method-derived ESI to the existing LDI and retires the previously active ESI, per operator policy. |
+| **Revocation / Termination** | MIS invalidates the LDI when keys are compromised, the device is decommissioned, or policy mandates retirement. Once revoked/terminated, an LDI **MUST NOT** be re-issued. |
+
+The MIS **MUST** maintain an authoritative mapping of ESI to LDI within the Trust Domain and **MUST NOT** allow duplicate or conflicting bindings.
+
+> **Logical Device Identity Lifecycle (Informative)**
+>
+> ```mermaid
```
phil-abb: This mermaid diagram is not rendering correctly

proposals/margo-identity-and-authorization-framework.md
```diff
@@ -0,0 +1,2563 @@
...
+
+Renewal is authenticated **directly with an existing SVID**:
+the client either presents its current X.509 SVID as a TLS client certificate (mTLS), or presents a JWT SVID as an HTTP Bearer token, and the MIS issues a new SVID for the same SPIFFE ID.
+
+> **Note:**
+> This endpoint **MAY** accept JWT SVID bearer authentication because the caller is already presenting an issued identity and requesting refreshed credentials for that same SPIFFE ID. The [JWT SVID Exchange Endpoint](#jwt-svid-exchange-endpoint) intentionally does **not** accept JWT SVID bearer authentication because it is the mechanism used to obtain a JWT SVID from proof tied to an existing X.509 SVID.
+
+| Item | Value |
+| :--- | :---- |
+| **Endpoint** | `POST /api/v1/identities/{spiffeIdEncoded}/renewals` |
```
phil-abb: Should this endpoint be "renewal" instead of "renewals"? 

Based on my understanding, you will only be renewing one item, so "renewal" may be more appropriate. But maybe I'm misunderstanding something.

proposals/margo-identity-and-authorization-framework.md
```diff
@@ -0,0 +1,2563 @@
...
+
+#### Terms introduced by this SUP <!-- omit from toc -->
+
+The following terms are defined by this SUP. They represent Margo-specific concepts that build on the SPIFFE primitives above.
+
+##### Policy-Based Authorization <!-- omit from toc -->
+
+Authorization based on verified **SPIFFE IDs** and associated attributes, evaluated locally within the Trust Domain - not on external token scopes.
+
+##### Margo Identity Service (MIS) <!-- omit from toc -->
```
phil-abb: I have a concern about ownership here. The MIS is an implementation that has to be running within the trust domain, but there is no clear expectation that anyone is going to "own" this service since it's meant to be a general service. So, the WFM and DFM vendors can assume someone else is going to provide it, so neither has it, and the customer needs to provide it. Or they both assume they need to provide it, so you end up with two implementations, etc. 

Its purpose makes sense in the context of the SUP, but I think we need to figure out the expectations on this. I don't think it makes sense for this to be an official Margo-supported implementation (I guess a SC decision). I think it probably makes sense to include a minimal implementation of this with the reference implementation, but that is different than Margo providing something that will be supported in any production scenarios.

proposals/margo-identity-and-authorization-framework.md
matlec: You're right, good catch! The impact is unclear because the SUP is actually under-specified. A leaf-only TLS chain + root-only Trust Bundle would break path building for verifiers. I'll fix this at the presenter side rather than make the Trust Bundle carry the intermediates, i.e.:
1. mTLS presenters MUST send the full SVID chain (incl. intermediates)
2. The `client_assertion` JWT used at the JWT SVID Exchange endpoint MUST include `x5c` with the full certificate chain
3. Verifiers MUST validate the presented certificate chain against the Trust Bundle's root certificates

With this in place, we can safely remove the MAY statements about intermediate CA certificates to avoid further confusion. Devices, WFMs, DFMs always receive intermediates via the SVID chain (in a TLS handshake, via `x5c`, or PEM chain), and only ever need root CA certificates in the Trust Bundle.

I'll push this change (incl. change to the PoC) tomorrow.

matlec: This is the relevant [commit](https://github.com/margo/specification-enhancements/pull/38/commits/46b5c7013487654485133cfb613f539671520620).

proposals/margo-identity-and-authorization-framework.md
matlec: Thanks, fixed!

proposals/margo-identity-and-authorization-framework.md
matlec: I agree, `/renewal` is the better option. The endpoint isn't really a REST collection, but rather an action on an identity. The singular option also matches the existing `/jwt-svid` endpoint.

proposals/margo-identity-and-authorization-framework.md
matlec: I share your concerns. The SUP (deliberately) defines MIS as a *role*, but that doesn't really resolve the deployment gap you're describing :) The WFM/DFM layer seems the right place for this to live: MIS is the identity plane that devices need before they can talk to anything, and the WFM/DFMs are the management planes devices and clients register with. But nothing in the SUP mandates this - and I also don't think this decision should be buried in a SUP.

Whether Margo should provide a production-quality open-source reference implementation of MIS - similar to how CNCF provides SPIRE for SPIFFE - is a question that belongs to the SC. My personal view: SPIRE is arguably why SPIFFE succeeded in practice, and a prod-quality Margo-maintained implementation would have the same effect on adoption.

proposals/miaf-poc/docker-compose.yml
```diff
@@ -0,0 +1,234 @@
...
+  # Step CA is an online CA for secure, automated certificate management. In this PoC,
+  # Step CA manages the root CA used by any X.509 SVIDs issued in this PoC.
+  step-ca:
+    image: smallstep/step-ca:latest
+    user: "0"
+    entrypoint: ["/bin/bash", "/etc/step-ca/entrypoint.sh"]
+    command:
+      - /usr/local/bin/step-ca
+      - --password-file
+      - /home/step/secrets/password
```
phil-abb, line 11: Is there something I need to do to create this password file before trying to run it? I didn't notice anything in the readme, but I may have missed something.

I'm getting the following error when the Docker Compose file tries to start up: 

`error reading /etc/step/password: error reading "/etc/step/password": is a directory`

```bash
mis-intermediate-bootstrap-1  | Issuing MIS intermediate offline using Step CA's ROOT as signer...
Container margo-identity-and-authorization-framework-poc-mis-intermediate-bootstrap-1 Waiting 
mis-intermediate-bootstrap-1  | error reading /etc/step/password: error reading "/etc/step/password": is a directory
mis-intermediate-bootstrap-1 exited with code 1
Container margo-identity-and-authorization-framework-poc-mis-intermediate-bootstrap-1 Error service "mis-intermediate-bootstrap" didn't complete successfully: exit 1
service "mis-intermediate-bootstrap" didn't complete successfully: exit 1
```

matlec: @phil-abb Thanks, I still had this password file locally. With the latest commit, this password file is generated by the Makefile.

To recover from this error, you might need to remove the `config/step-ca/password` directory that has been created by Docker:
```
make down
rm -rf config/step-ca/password
make up
```

proposals/margo-identity-and-authorization-framework.md
samuel-greising: I agree with @matlec pov on the adoption! The comparison with SPIRE is just perfect. I can offer testing an early solution in an enterprise grade environment with WFM & DFM combination.

phil-abb:
@margo/technical-wg This SUP Margo identity and authorization framework (MIAF) is ready for final review. Please review the proposal and add any questions/comments to the PR.

phil-abb:
@matlec We had an internal discussion about this today and have a question around the intention of putting `/margo/` in the SPIFFE ID. 

>Path namespace. SPIFFE IDs issued under a MIAF identity profile MUST have a path beginning with /margo/. Each identity profile claims a non-conflicting sub-prefix and defines its structure (path conventions for WFMs and WFM Clients are in the [Margo WFM Identity Profile](https://github.com/margo/specification-enhancements/blob/feat/miaf-wfm-identity-sup/proposals/wfm-identity-profile.md)). To keep /margo/ a reliable signal of MIAF provenance, non-MIAF SVIDs in the same Trust Domain MUST NOT use it.

Is there a reason for wanting to include the `/margo/` prefix other than collision protection?

There are some concerns about binding the ID to a specification instead of the workload. While it's true that the device and WFM are implementing the Margo specification, do we really need to include `/margo/` in the SPIFFE ID because it doesn't really address the question of "What is this workload?" Also, it's coupling identity to the spec directly, which might not be ideal, especially if you're looking to use the SPIRE infrastructure for more than just Margo stuff. Also, it is leaking implementation details, which probably shouldn't be in the SPIFFE ID.

matlec:
@phil-abb Fair points... If I understand correctly, I think your main point concern really is:

> binding the ID to a specification instead of the workload

And you're right on the principle. In SPIFFE the ID is just a name. What a workload *is* gets established when its SVID is issued, not by the path. SPIFFE leaves the attestation mechanism to the implementation. In SPIRE this would be workload attestation and the registration entry that maps the attested workload to an ID.

The reason it's in the path anyway comes down to one constraint: the SPIFFE ID containing the trust domain and the path is the only part of the identity a *remote* verifier can see. When one vendor's WFM validates an mTLS connection from another vendor's WFM client, it has the SVID and nothing else (e.g., no access to the other side's selectors or local metadata). Recognizing each other across vendors by construction is the whole point of MIAF, so that signal has to live in the SVID. If we take `/margo/` out, recognition falls back to allowlists every operator maintains by hand.

So I'd reframe `/margo/` as a namespace claim, closer to a DNS domain or a Kubernetes namespace than to "implements the Margo spec." That's also why I don't think it leaks implementation details: versions and hostnames are the stuff that shouldn't be in an ID, and a namespace isn't that.

With the current proposal in MIAF, the coupling stays narrow. The operator decides on the trust domain, the `<wfm-id>` and `<wfm-client-id>`. Margo only claims one labeled subtree, only for components that are Margo by definition.

This is all based on the assumption that an operator's trust domain can host both Margo (control-plane) components and other workloads side by side. Given that, the `/margo/` namespace is what keeps both cleanly separated. If every Margo deployment instead got its own dedicated trust domain, the prefix would be largely redundant. But that's not the model I've been working from.

On the practical side, this schema needs no custom SPIRE. A registration entry takes whatever SPIFFE ID you hand it. One SPIRE and one trust domain can serve Margo and non-Margo workloads side by side. SPIRE doesn't care what's in the path, and `/margo/` IDs coexist with `/ns/<ns>/sa/<sa>` or anything else with no conflict. The only prefix SPIRE reserves for itself is `/spire/` (for its own server and agents), so `/margo/` just sits next to it as another namespace.

Does this address the concerns from your discussion?

phil-abb:
>Does this address the concerns from your discussion?

@matlec - Thanks. The explanation makes sense. I'll pass the information along to the people who raised the question internally and let you know if there are any follow-up questions/concerns.

phil-abb:
Merging in this PR since the SUP has been approved: https://github.com/margo/specification-enhancements/issues/70#issuecomment-4768118798

PR #40, opened 2026-02-05
[DRAFT] Move Application Deployment Templating to the WFM

jjaswanson4:
This SUP is to adjust the features and functionality of the agent deployed to devices under the management of a fleet/workload manager - moving more of the “heavy lifting” up to the manager level, while keeping the agent feature set more streamlined.

Specifically, this SUP moves templating of the application deployment (and the corresponding tooling) to the WFM level, instead of expecting the agent/device to handle it.


phil-abb:
@jjaswanson4 Can you please mark this as a draft PR until it's ready to be submitted?

jjaswanson4:
> @jjaswanson4 Can you please mark this as a draft PR until it's ready to be submitted?

Updated title

jjaswanson4:
> Is the intention for this SUP to indicate that manifests MUST be rendered by the WFM and only provide rendered manifests to the device? Or is this optional?
This should be required, as distributing/maintaining the helm binaries on every device (in an agent package or otherwise) is not the intention of helm, nor a best practice.
> If it's required, what changes to the specification are being proposed
The WFM needs to take on two main "new" responsibilities:
1. Rendering the deployment using device information reported from the agent on the devices (so device-specific vars/etc can be used to render the deployment)
2. The WFM must provide an endpoint where the agent on the device can retrieve the rendered deployment

> Is the intention of this SUP to only do this for Helm Charts? Or all deployment profile types?
If this intention is to do this for all deployment types, then I would like to see the details of how this is going to be handled in the SUP; if it's only for Helm, then it's more self-explanatory

I think it's fine to start with "helm only" - personally, I think we should be doing this for all deployment types, but the helm example is the most glaring, in my opinion.

> What changes, if any, are you proposing be made to the Application Description to support this alternative approach?

None

> What changes are you proposing be made to the desired state, since the desired state would be different from what is currently defined

Desired state would essentially equal the rendered application deployment - we'll need to outline how we want to represent drift or deltas

> Can you provide references for this quote showing that this is not a generally accepted best practice?

Review any of the reference architectures for ArgoCD, or helm's expected consumption, or how Ansible handles runtime variable rendering.

For example: https://argo-cd.readthedocs.io/en/stable/operator-manual/architecture/

Our use case is aligned to the "deploy" portion of that architecture.

> The SUP indicates the following. If this is the case, I think the SUP needs to indicate the specification changes required so that the WFM would have all this content, because right now the spec assumes this is coming from somewhere else
> Gathering required assets (container images, etc) from the manager

I have a second SUP that more clearly defines having the manager be a "cache" or "pull-through" for required content. In reality however, with this approach, the amount of content the device needs to pull directly is lowered - it does not need to contact a code repo, for example, it only needs the application deployment definition and the corresponding container images.

> It's not clear just from the diagram in the Technical Proposal section what changes to the SUP are being proposed. A few specific questions about the diagram:
> What is being proposed for the interactions with the image registry

Right now, there is no change from existing behavior (see my above comment about how a second SUP should add cache/pull-through to the WFM)

> What is being proposed for adding support for Git

This SUP is specific to how application deployments should be rendered, and where - I do think we support having helm charts/etc be available via OCI or git, and the manager can pull from either seamlessly when onboarding a new application to be deployed to devices, but that sounds like another SUP.

> What is being proposed for changes to the application deployment definition

The deployment should be the exact rendered "thing for the device to do" - the agent simply downloads it and applies it to the corresponding endpoint

> This is related to this SUP, but what is being proposed for how to handle Helm charts using a feature requiring interactions with the Kubernetes APIs?

See the above architecture information - in a production-level setup, helm is used as a rendering tool, not as a deployment tool (even though helm can do that) - a higher level construct (like argo, or in our context, a WFM) handles using the tool to render the deployment

> If the intention is not to support these features, what are the proposed alternative solutions? For example, if a Helm chart is using a lookup, what is being proposed as an alternative to get the same information from the device?

The WFM should be able to use information reported by the device (which we already have some concepts around - reporting CPU, memory, devices, whatever) when rendering the application deployment. This lowers the workload on the devices, and leverages that data we already intend to collect.

jjaswanson4:
Adding an example ApplicationDeployment object to be used to drive discussion:
```yaml
---
apiVersion: margo.org/v1-alpha1
kind: ApplicationDeployment
metadata:
  id: f9483655-2c7b-4155-8d82-5d5db4ce30e0
  name: hello-world-deployment
  description: A deployment of the hello-world application to device device1
  version: "1.0"
spec:
  # This should match  
  catalogItem: 
    name: hello-world
    version: 1.0.0
    # If WFM is generating UUIDs for catalog items
    uuid: 743e1fbd-1cda-403e-a3dc-634ef9f23099
  # What deployment profile was used to render the deployment
  usedDeploymentProfile:
    - type: helm.v3
      id: com-northstartida-hello-world-helm.v3-a
      components:
        - name: hello-world
          properties:  
            repository: oci://northstarida.azurecr.io/charts/hello-world
            revision: 1.0.1
            wait: true
  deployment: |
    # The rendered application deployment - example contains a few resources
    ---
    apiVersion: v1
    kind: Pod
    metadata:
      name: hello-world-pod
      ...
    ---
    apiVersion v1
    kind: Service
    metadata:
      name: hello-world-svc
      ...
  # What parameters were used with the rendering tool (such as helm) to render the deployment
  usedParameters:
    greeting:
      value: Hello
      targets:
      - pointer: global.config.appGreeting
        components: ["hello-world"]
    greetingAddressee:
      value: World
      targets:
      - pointer: global.config.appGreetingAddressee
        components: ["hello-world"]
```
This isn't intended to be 100% what the object is, just a starting point.

This would be rendered by the WFM, and simply retrieved by the end device and applied to the corresponding endpoint (docker, k8s, etc).

This is also *heavily* aligned to what's being proposed here: https://github.com/margo/specification-enhancements/pull/47

phil-abb:
```yaml
  deployment: |
    # The rendered application deployment - example contains a few resources
    ---
    apiVersion: v1
    kind: Pod
    metadata:
      name: hello-world-pod
      ...
    ---
    apiVersion v1
    kind: Service
    metadata:
      name: hello-world-svc
      ...
```

For big applications, the YAMLs from the rendered templates can be very large. Are there any concerns with putting the manifests inline like this? I guess it's all relative, since it's text, it would have to be really large for it to become a problem.

jjaswanson4:
> For big applications, the YAMLs from the rendered templates can be very large. Are there any concerns with putting the manifests inline like this? I guess it's all relative, since it's text, it would have to be really large for it to become a problem.

That's a fair question - since this is an object that's "owned" by Margo (it's an object specific to the standard and the software that conforms to it), I think it's fine to state "this can be very large given the size of the application being deployed".

You're correct - it's just text, so even if its thousands of lines, I don't see that being an issue.

ajcraig:
I wonder if it would be good during onboarding of the wfm client for there to be a negotiation between the client and the wfm on who renders the manifests.
 
I'd assume after this sup goes through, templating within the WFM would be an **optional** feature. I don't think we are making the assumption, wihtin this sup, every WFM can render every manifest. Who knows what we will need to support in this specification in the coming years.

Something like this during onboarding/exchange:

- Device reports what `supportedManifests` within reported capabilities
    - helm or compose
- WFM replies to client 
    - I will render helm/compose within WFM, you should expect rendered manifests(whether kubernetes or docker yamls + future examples).

Or, does the device just react to the ApplicationDeployment spec and apply whatever comes? 
@phil-abb @jjaswanson4  thoughts?

nilanjan-samajdar:
@jjaswanson4 , @phil-abb , @ajcraig 
The key question to this SUP is the element of "trust".
Does a Device Maker trust the WFM to render the `ApplicationDeployment`, without breaking it's trust rules for access to resources and it's control-plane for the runtime ?
- If yes, Devices can relegate rendering to the WFM
- If No, Devices would want to render the `ApplicationDeployment` based on it's design constraints as well as operational constraints.
In summary, the negotiation suggested by @ajcraig makes sense, but who has the final word ?

PR #46, opened 2026-03-26
SUP for supporting Helm V4

phil-abb:
SUP for adding support for Helm V4 to the specification.


chrisgclayton:
I agree with the expanded support for V4 etc. The question I have is the helm.v3 becoming helm leads me to question what happens when helm v5 comes along? Do we still have just helm and what if there are breaking changes etc. I wonder if splitting it to deployment type helm and a deployment type supported versions concept may make sense for forward compat?

phil-abb:
> I agree with the expanded support for V4 etc. The question I have is the helm.v3 becoming helm leads me to question what happens when helm v5 comes along? Do we still have just helm and what if there are breaking changes etc. I wonder if splitting it to deployment type helm and a deployment type supported versions concept may make sense for forward compat?

I'm not sure, but if anything, I think we'd want to add something that indicates the Chart API Version and not the Helm version. The way I see it is if Helm V5 continues to support Chart API v2, then we wouldn't need to make any changes since the docs would say we support Helm with Chart API v2.  We would need to make a change at some point once Helm starts working with Chart API v3. So, maybe once the Chart API v3 is released, we would need to add something additional.

phil-abb:
@margo/technical-wg - This SUP has been voted on and approved. https://github.com/margo/specification-enhancements/issues/51

PR #50, opened 2026-04-16
Margo Device Roles to Capabilities

ajcraig:


phil-abb:
@margo/technical-wg A new SUP has been started by @ajcraig. If anyone is interested in helping work on fleshing out the details of this SUP, please contact @ajcraig

ajcraig:
SUP Updated with technical details section. 

Ready for an initial review / refinement!

proposals/sup_device_roles_to_capabilities.md
```diff
@@ -0,0 +1,287 @@
...
+
+## Device-Specific Runtimes (active CURRENT SUP)
+
+Devices MUST report vendor-specific runtimes and their named capabilities using the `vendorRuntimes` field in the `capabilities` object. Application description files may specify required vendor runtimes and capabilities. The WFM MUST validate that a device has the required vendor runtime and capabilities before dispatching a deployment.
+
+---
+
+## Gateway and Compound Devices(APPROVED SUP)
+
+Gateway devices MUST report their capabilities to the WFM as a single `DeviceCapabilitiesManifest`, reflecting the aggregate capabilities of all sub-devices they manage. Gateway operational modes:
```
phil-abb: This is only correct for the [opaque gateway](https://github.com/margo/specification-enhancements/blob/main/proposals/gateway.md#opaque-gateway) but not the [see-thru gateway](https://github.com/margo/specification-enhancements/blob/main/proposals/gateway.md#see-thru-gateway)

proposals/sup_device_roles_to_capabilities.md
```diff
@@ -0,0 +1,287 @@
...
+
+* **`supportedRuntimes`** — for well-known, Margo-standard OCI runtimes (docker, podman, kubernetes).
+* **`vendorRuntimes`** — for vendor-specific runtime environments that extend or replace the standard runtimes. A device MAY report only `vendorRuntimes` if none of the standard `SupportedRuntimeType` values apply.
+
+A device MUST report at least one entry across `supportedRuntimes` or `vendorRuntimes`.
+
+| Attribute         | Type               | Required? | Description |
+|-------------------|--------------------|-----------|-------------|
+| wfmClient         | WfmClient          | Y         | Reports the WFM client present on the device. See [WfmClient Attributes](#wfmclient-attributes). |
+| otelCollector     | OtelCollector      | Y         | Reports the OTEL collector present on the device. See [OtelCollector Attributes](#otelcollector-attributes). |
```
phil-abb: Is there a reason these are "object" types instead of just boolean types? Is there an intention of having other properties associated with "wfmClient" and "otelCollector" other than just "present"? If so, this might be OK, but it should be commented on. If not, then I think these should just be booleans instead of objects.

proposals/sup_device_roles_to_capabilities.md
```diff
@@ -0,0 +1,287 @@
...
+| supportedRuntimes | []SupportedRuntime | N*        | Standard Margo OCI runtimes available on the device. See [SupportedRuntime Attributes](#supportedruntime-attributes). Required if no `vendorRuntimes` are reported. |
+| supportedManifests| []SupportedManifestType | Y    | Manifest/deployment formats the device can receive and process locally. See [SupportedManifestType](#supportedmanifesttype). |
+| vendorRuntimes    | []VendorRuntime    | N*        | Vendor-specific runtimes, used as an alternative or supplement to `supportedRuntimes` when standard runtime types do not apply. See [VendorRuntime Attributes](#vendorruntime-attributes). Required if no `supportedRuntimes` are reported. |
```
phil-abb: It's not clear to me why we need a seperate "vendorRuntimes" object. We are using the term "runtime" in several different places now, and it's not very clear what we mean in each place. There doesn't seem to be consistency in what we're referring to. Here, the runtime seems to be the software the deployment runs on (so something other than Docker, Podman, or Kubernetes). If this is the case, and we want to allow an app vendor and a device vendor to team up to deploy something like a virtual machine image, do we really need all those properties under "vendorRuntimes"? 

Instead, could we have a runtime enum with Docker, Podman, Kubernetes, Other,  and if "Other" is selected, they provide the name in a special name property? And the same for manifests with Helm, Compose, and others.

What is it we're trying to accomplish with these vendorRuntime properties? What is the use case(s) we're trying to solve? How does this additional information help to solve it? Because this vendor runtime information seems to go beyond the software used to run a workload and the type of workload it is, I'm getting confused by what we want to accomplish with this.

```yaml
"runtimeRef": "com.acme.special-rt",
                    "vendor": "Acme Corp",
                    "vendorId": "com.acme",
                    "modelNumberId": "com.acme.edge-devices.X1000",
                    "version": "3.1.0",
                    "runtimeCapabilities": [
                        "foo",
                        "bar"
                    ],
                    "properties": {
                        "osVersion": "Debian 12"
                    }
```

proposals/sup_device_roles_to_capabilities.md
ajcraig: Addressed in latest commit

proposals/sup_device_roles_to_capabilities.md
ajcraig: Addressed in latest commit

proposals/sup_device_roles_to_capabilities.md
ajcraig: Tried to remove all mentions of this injection. Will wait on the other SUP to make it's way through the process before we integrate.

proposals/sup_device_roles_to_capabilities.md
phil-abb: The example needs to be updated because it still has the old structure

This is what it should be now:
```json
"capabilities": {
            "wfmClient":  true,
            "otelCollector": true,
            "supportedRuntimes": [
                { "type": "kubernetes" },
                { "type": "podman" }
            ],
            "supportedManifests": [
                "helm",
                "compose"
            ],
       }
```
            
Maybe the same type of change should be considered for supportRuntimes.type as well. If we're not going to have other properties, we don't really need it. So, it could be:

```json
"capabilities": {
            "wfmClient":  true,
            "otelCollector": true,
            "supportedRuntimes": [
                "kubernetes",
                "podman" 
            ],
            "supportedManifests": [
                "helm",
                "compose"
            ],
       }
```

ajcraig: should be good to go

phil-abb:
@margo/technical-wg This SUP to remove device roles from the device capabilities is ready for final review. Please review the proposal and add any questions/comments to the PR.

phil-abb:
I'm going to challenge the need for defining the `supportedRuntimes`

```json
"supportedRuntimes": [
  "kubernetes",
  "podman"
],
```

Why do we feel we need to define this? Is it just information, or is there a purpose behind calling this out specifically? If it's informational, that's fine, and this array should allow any runtime and not just Kubernetes, Podman, and/or Docker.

If we want to put meaning behind it, does it matter to an app vendor if it's Kubernetes, Docker, or Podman when they are indicating their application supports either Helm or Compose? Should app vendors be required to specify in the deploymentProfile section for their Compose applications if it's to be runnable on Podman, Docker, or both? It shouldn't matter.

If it's here because we want to support "custom runtimes," then I think there is probably a better way to handle this in general.

julienduquesnay-se:
I don't think there is a need to categorize capabilities (supportedRuntimes, supportedManifests). A flat list with is enough and will be easier to maintain (no need to create new categories and to make a choice is a capability could be part of multiple categories for example). A flat list is enough for the WFM to make scheduling decisions or provide support is workload assignment to the end-user.
Also, I think capabilities could be listed as an array of simple strings, no need to have a key/value. If the string is present the device supports the capability, if it is not present the device does not support the capability.

julienduquesnay-se:
I think we should allow device vendor to advertise 'non-standard" capabilities. This could be done by using a prefix like 'vsv.' (for vendor specific value) to identify the non-standard values. If we are concerned about value collision between vendor, we can also add a vendor prefix, but this adds more overhead to assign maintain that list of prefixes.

phil-abb:
> I think we should allow device vendor to advertise 'non-standard" capabilities. This could be done by using a prefix like 'vsv.' (for vendor specific value) to identify the non-standard values. If we are concerned about value collision between vendor, we can also add a vendor prefix, but this adds more overhead to assign maintain that list of prefixes.

@julienduquesnay-se that's what this SUP is defining: https://github.com/margo/specification-enhancements/pull/56

proposals/sup_device_roles_to_capabilities.md
```diff
@@ -0,0 +1,241 @@
...
+
+Devices MUST provide the Workload Fleet Management service with both their resources and their capabilities. Resources are quantitative attributes (CPU, memory, storage) used for workload scheduling. Capabilities describe what the device can do — which runtimes it supports, and which manifest types it can process.
+
+This proposal replaces the `roles` field in the current `DeviceCapabilitiesManifest` with a `capabilities` object, while keeping the `resources` object unchanged.
+
+## Requirements Changes
+
+The `roles` field in the current `DeviceCapabilitiesManifest` `properties` schema is replaced by a `capabilities` object. All references to static device roles (Standalone Cluster, Cluster Leader, Standalone Device) in the specification will be updated following this SUP.
+
+All previously hard-coded device requirements (WFM client, OCI runtime, OTEL collector) are now expressed as capabilities within the `capabilities` object. The WFM client and OTEL collector are reported as boolean flags, while OCI runtime support is captured through `supportedRuntimes`. A device MUST report at least one entry across `supportedRuntimes`, and MUST report at least one `supportedManifest`.
```
phil-abb: Should `supportedManifest` be `supportedDeploymentTypes` instead, so that it matches what is in the application definition?

ajcraig:
> I'm going to challenge the need for defining the `supportedRuntimes`
> 
> ```json
> "supportedRuntimes": [
>   "kubernetes",
>   "podman"
> ],
> ```
> 
> Why do we feel we need to define this? Is it just information, or is there a purpose behind calling this out specifically? If it's informational, that's fine, and this array should allow any runtime and not just Kubernetes, Podman, and/or Docker.
> 
> If we want to put meaning behind it, does it matter to an app vendor if it's Kubernetes, Docker, or Podman when they are indicating their application supports either Helm or Compose? Should app vendors be required to specify in the deploymentProfile section for their Compose applications if it's to be runnable on Podman, Docker, or both? It shouldn't matter.
> 
> If it's here because we want to support "custom runtimes," then I think there is probably a better way to handle this in general.

This has been changed to simply `oci` since that covers the runtimes I described.

nseba:
The SUP attributes reference several attribute definitions, but none of them are defined in this SUP. Are these subject to follow-ups? 

Does the WFM client have any need for more attributes? What purpose would it serve?

> 
| Attribute         | Type               | Required? | Description |
|-------------------|--------------------|-----------|-------------|
| wfmClient         | boolean         | Y         | Reports the WFM client present on the device. See [WfmClient Attributes](#wfmclient-attributes). |
| otelCollector     | boolean     | Y         | Reports the OTEL collector present on the device. See [OtelCollector Attributes](#otelcollector-attributes). |
| supportedRuntimes | array | Y        | Standard Margo OCI runtimes available on the device. See [SupportedRuntimes](#supportedruntime). |
| supportedDeploymentTypess| array | Y    | Manifest/deployment formats the device can receive and process locally. See [supportedDeploymentTypes](#supportedDeploymentTypes). |

ajcraig:
> The SUP attributes reference several attribute definitions, but none of them are defined in this SUP. Are these subject to follow-ups?
> 
> Does the WFM client have any need for more attributes? What purpose would it serve?
> 
> > 
> 
> Attribute	Type	Required?	Description
> wfmClient	boolean	Y	Reports the WFM client present on the device. See [WfmClient Attributes](#wfmclient-attributes).
> otelCollector	boolean	Y	Reports the OTEL collector present on the device. See [OtelCollector Attributes](#otelcollector-attributes).
> supportedRuntimes	array	Y	Standard Margo OCI runtimes available on the device. See [SupportedRuntimes](#supportedruntime).
> supportedDeploymentTypess	array	Y	Manifest/deployment formats the device can receive and process locally. See [supportedDeploymentTypes](#supportedDeploymentTypes).

Good catch, those references have been removed.

ajcraig:
@julienduquesnay-se 
> I don't think there is a need to categorize capabilities (supportedRuntimes, supportedManifests). A flat list with is enough and will be easier to maintain (no need to create new categories and to make a choice is a capability could be part of multiple categories for example). A flat list is enough for the WFM to make scheduling decisions or provide support is workload assignment to the end-user. Also, I think capabilities could be listed as an array of simple strings, no need to have a key/value. If the string is present the device supports the capability, if it is not present the device does not support the capability.

Couple points on your comments:

- Could you provide an example of how that would look? IMO, I think as we move forward the desire to make these items more robust just grows. I would fear a flat list would need redone if more verbosity is required. But open to see your proposal. 
- I'm not sure which of the items I added would be part of two different categories. My goal was to breakdown what the roles were covering previously, which included WFM client, Otel collector, manifest templating tool(compose or helm), and the oci runtime.

julienduquesnay-se:
> @julienduquesnay-se
> 
> > I don't think there is a need to categorize capabilities (supportedRuntimes, supportedManifests). A flat list with is enough and will be easier to maintain (no need to create new categories and to make a choice is a capability could be part of multiple categories for example). A flat list is enough for the WFM to make scheduling decisions or provide support is workload assignment to the end-user. Also, I think capabilities could be listed as an array of simple strings, no need to have a key/value. If the string is present the device supports the capability, if it is not present the device does not support the capability.
> 
> Couple points on your comments:
> 
> * Could you provide an example of how that would look? IMO, I think as we move forward the desire to make these items more robust just grows. I would fear a flat list would need redone if more verbosity is required. But open to see your proposal.
> * I'm not sure which of the items I added would be part of two different categories. My goal was to breakdown what the roles were covering previously, which included WFM client, Otel collector, manifest templating tool(compose or helm), and the oci runtime.

@ajcraig 

I would make the new `capabilities` attribute an array of strings instead of an object. Each string in that array identifies a capability supported by the device.
A payload could look like:
```
{
    "properties": {
        "id": "northstarida.xtapro.k8s.edge",
        "vendor": "Northstar Industrial Devices",
        "modelNumber": "332ANZE1-N1",
        "serialNumber": "PF45343-AA",
        "resources": {
            ...
        },
        "capabilities": [
            "wfmClient",
            "otelCollector",
            "oci-runtime",
            "compose",
        ]
    }
}
```
Adding new capabilities (runtime, deployment type, service, etc.) in the future would only require specifying one meaningful name and not to either match an existing category or create a new one (here I consider `supportedRuntime` and `supportedDeploymentTypes` to be categories).

phil-abb:
> @ajcraig
> 
> I would make the new `capabilities` attribute an array of strings instead of an object. Each string in that array identifies a capability supported by the device. A payload could look like:
> 
> ```
> {
>     "apiVersion": "device.margo.org/v1alpha1",
>     "kind": "DeviceCapabilitiesManifest",
>     "properties": {
>         "id": "northstarida.xtapro.k8s.edge",
>         "vendor": "Northstar Industrial Devices",
>         "modelNumber": "332ANZE1-N1",
>         "serialNumber": "PF45343-AA",
>         "resources": {
>             ...
>         },
>         "capabilities": [
>             "wfmClient",
>             "otelCollector",
>             "oci-runtime",
>             "compose",
>         ]
>     }
> }
> ```
> 
> Adding new capabilities (runtime, deployment type, service, etc.) in the future would only require specifying one meaningful name and not to either match an existing category or create a new one (here I consider `supportedRuntime` and `supportedDeploymentTypes` to be categories).

I'm not really a fan of this. 

I could potentially see changing this: 
```json
 "capabilities": {
            "wfmClient": true,
            "otelCollector": true,
            "supportedRuntimes": [
                "oci"
            ],
            "supportedDeploymentTypes": [
                "helm",
                "compose"
            ]
        }
```

To this:
```json
 "properties": {
            ...
            "wfmClient": true,
            "otelCollector": true,
            "supportedRuntimes": [
                "oci"
            ],
            "supportedDeploymentTypes": [
                "helm",
                "compose"
            ]
        }
```

But I don't like the idea of `capabilities` just being a free-form list of anything people want to put in it. I think anything we put under "properties" should be specifically defined by Margo and have intentional meaning.

nilanjan-samajdar:
I just added this comment in the SUP for vendor extensions - https://github.com/margo/specification-enhancements/pull/56
I realize now, that it makes more sense to first put it here as an extension to what's defined above.

---

May I suggest the following representation for "Device Capabilities" and "Vendor defined Resources" - 
1. Have a flexible "Dictionary" of vendor specific resources / cpabilities which can be updated by a Device Vendor at the time a new class of devices is introduced to the WFM, maybe offline or during device enrollment / onboarding.
2. The WFM can treat these vendor specific resources as long as they know how to 'count' the resources, the vendor provides this information as well as part of the dictionary.

Organization of data ad relationships
1. A class called 'DeviceCapabilitiesManifest' which has an attribute - 'properties'. 
* This 'properties' attribute consists of - "id", "vendor", "modelNumber", "serialNumber", "roles", "resources" and "Vendor Defined Resources".  
* The "Vendor defined Device Resources" is further defined below.
* The "resources" attribute further consists of "cpu", "storage", "peripherals", "interfaces" and "Device Capabilities". The "Device Capabilities" is realized by a separate class as given below.
2. A class called 'Device Capabilities', which is an array of of enumerated strings realized by a class 'Device Capability Dictionary' defined below
3. A class called 'Vendor defined Device Resources', which is a key-value pair, where the key comes from the 'resource-name' attribute in a class called 'Device Resource Dictionary'.
4. The Device Resource Dictionary has the following members - 'resource-name', 'vendor', 'measurement unit'

```mermaid
classDiagram

%% -------------------------------
%% Root Class
%% -------------------------------
class DeviceCapabilitiesManifest {
  +id
  +vendor
  +modelNumber
  +serialNumber
  +roles
  +resources
  +vendorDefinedResources
}

%% -------------------------------
%% Resources Block
%% -------------------------------
class Resources {
  +cpu
  +storage
  +peripherals
  +interfaces
  +deviceCapabilities
}

DeviceCapabilitiesManifest --> Resources : contains

%% -------------------------------
%% Device Capabilities
%% -------------------------------
class DeviceCapabilities {
  +capabilities : enum[]
}

Resources --> DeviceCapabilities : uses

%% -------------------------------
%% Capability Dictionary (Enum)
%% -------------------------------
class DeviceCapabilityDictionary {
  <<enumeration>>
  CapabilityA
  CapabilityB
  CapabilityC
}

DeviceCapabilities --> DeviceCapabilityDictionary : values from

%% -------------------------------
%% Vendor Defined Resources
%% -------------------------------
class VendorDefinedDeviceResources {
  +resources : key-value
}

DeviceCapabilitiesManifest --> VendorDefinedDeviceResources : contains

%% -------------------------------
%% Resource Dictionary
%% -------------------------------
class DeviceResourceDictionary {
  +resourceName
  +vendor
  +measurementUnit
}

VendorDefinedDeviceResources --> DeviceResourceDictionary : key reference
```

phil-abb:
> I'm not really a fan of this.
> 
> I could potentially see changing this:
> 
> ```json
>  "capabilities": {
>             "wfmClient": true,
>             "otelCollector": true,
>             "supportedRuntimes": [
>                 "oci"
>             ],
>             "supportedDeploymentTypes": [
>                 "helm",
>                 "compose"
>             ]
>         }
> ```
> 
> To this:
> 
> ```json
>  "properties": {
>             ...
>             "wfmClient": true,
>             "otelCollector": true,
>             "supportedRuntimes": [
>                 "oci"
>             ],
>             "supportedDeploymentTypes": [
>                 "helm",
>                 "compose"
>             ]
>         }
> ```

Also, since the document itself is called `DeviceCapabilitiesManifest` do we need a property called "capabilities"? Isn't this already implied by the name of the document?

matlec:
Seeing `wfmClient` as part of the `capabilities` object leaves me puzzled about the relationship between the WFM client that reports the capability and the "device" for which the capability is reported.

The SUP defines `wfmClient: boolean` simply as "Reports the WFM client present on the device.". But the WFM client is the entity that actually reports the capabilities, so the field doesn't seem to carry information the WFM doesn't already have. Walking through the cases:

- **Single-node device (1:1):** `wfmClient: true` is implied by the fact that anything is reporting at all ;).
- **See-thru gateway, sub-device manifest:** `wfmClient: false` is implied by the nested URL path (`/clients/{clientId}/capabilities/{device-id}...`) - the gateway forwards on behalf of sub-devices that don't run their own WFM client.
- **Opaque gateway, aggregate manifest:** the gateway has a WFM client, the sub-devices don't - a single boolean can only represent the gateway which by definition must run a WFM client.

In every case `wfmClient` is either redundant or ambiguous. What is it meant to indicate?

There is actually another case I'd like to raise: if we think about multi-node Kubernetes clusters (as envisioned [here](https://docs.margo.org/concepts/edge-compute-devices/devices)), how do we expect a WFM client managing workloads on a multi-node Kubernetes cluster to report device capabilities once `roles` is gone? One manifest for the cluster as a whole? One per node, with `wfmClient: true` only on the node where the WFM client actually runs?

I'm not trying to resolve device-identity / "WFM client to device binding" inside this SUP, but I'd like us to start the conversation, because the choices we make here constrain it. I'd suggest either (a) dropping `wfmClient` until it carries information beyond restating its own existence, or (b) including explicit text on what `wfmClient: true` / `false` means in the gateway and multi-node cluster cases so we don't bake in an assumption we'll have to undo.

ajcraig:
Good comment, @matlec. 

To get this through, I'm open to remove the `wfmClient` boolean, while enabling extensions later such as gateway/multi-node cluster. 
- I can even think about a client that is replicated across 2 nodes.

phil-abb:
> Also, since the document itself is called `DeviceCapabilitiesManifest` do we need a property called "capabilities"? Isn't this already implied by the name of the document?

@ajcraig 

Also, I don't think we need the "resources" category either. With the way we have it now, we have some capabilities under "resources" and some under "capabilities". I think we should remove both "resources" and "capabilities". Let me know what you think. If you don't agree, I'll create an alternative SUP with the proposal.

So, I think we should have this:

```json
{
    "properties": {
        "id": "northstarida.xtapro.k8s.edge",
        "vendor": "Northstar Industrial Devices",
        "modelNumber": "332ANZE1-N1",
        "serialNumber": "PF45343-AA",
        "cpu": [
            {
                "cores": 24,
                "architecture": "x86_64"
            }
        ],
        "memory": "59 Gi",
        "storage": "1862 Gi",
        "peripherals": [
            {
                "type": "gpu",
                "manufacturer": "NVIDIA"
            }
        ],
        "interfaces": [
            {
                "type": "ethernet"
            },
            {
                "type": "wifi"
            }
        ],
        "otelCollector": true,
        "supportedRuntimes": [
            "oci"
        ],
        "supportedDeploymentTypes": [
            "helm",
            "compose"
        ]
    }
}
```

ajcraig:
@phil-abb 
> So, I think we should have this:
Agreed on what you proposed. 

Proposal has been updated via my last commit to reflect that more flat / concise structure. Let me know if you find any issues with the edits I made....it's late in the week :)

proposals/sup_device_roles_to_capabilities.md
```diff
@@ -0,0 +1,222 @@
...
+### Properties Attributes
+
+| Field                    | Type              | Required? | Description |
+|--------------------------|-------------------|-----------|-------------|
+| id                       | string            | Y         | Unique deviceID assigned to the device via the Device Owner. |
+| vendor                   | string            | Y         | Defines the device vendor. |
+| modelNumber              | string            | Y         | Defines the model number of the device. |
+| serialNumber             | string            | Y         | Defines the serial number of the device. |
+| cpu                      | array             | Y         | List of CPU entries available on the device. Each entry includes `cores` (integer) and `architecture` (see [CpuArchitectureType](#cpuarchitecturetype)). |
+| memory                   | string            | Y         | Total memory available on the device (e.g. `59 Gi`). |
```
singhmj-1, line 113: If the SUP allows then I would suggest splitting the memory input into two fields: an integer `memory` field and a `memoryUnit` enum (e.g., Mi, Gi). This avoids string parsing and simplifies back-end validation.

proposals/sup_device_roles_to_capabilities.md
```diff
@@ -0,0 +1,222 @@
...
+| storage                  | string            | Y         | Total storage available on the device (e.g. `1862 Gi`). |
```
singhmj-1, line 114: If the SUP allows then I would suggest splitting the storage input into two fields: an integer `storage` field and a `storageUnit` enum (e.g., Mi, Gi).

proposals/sup_device_roles_to_capabilities.md
```diff
@@ -0,0 +1,222 @@
...
+    "apiVersion": "device.margo.org/v1alpha1",
+    "kind": "DeviceCapabilitiesManifest",
+    "properties": {
+        "id": "northstarida.xtapro.k8s.edge",
+        "vendor": "Northstar Industrial Devices",
+        "modelNumber": "332ANZE1-N1",
+        "serialNumber": "PF45343-AA",
+        "cpu": [
+            {
+                "cores": 24,
```
singhmj-1, line 185: If the SUP allows then I would suggest splitting the cpu input into two fields: an integer cpu field and a cpuUnit enum (e.g., core, millicore).

proposals/sup_device_roles_to_capabilities.md
```diff
@@ -0,0 +1,222 @@
...
+## Enumerations
+
+> Note: These enumerations can be expanded in the future through the SUP process. This SUP aims to set the framework for future expansion. 
+
+### CpuArchitectureType
+
+| Permissible Values | Description |
+|--------------------|-------------|
+| amd64              | AMD 64-bit architecture. |
+| x86_64             | x86 64-bit architecture. |
```
singhmj-1: `amd64` already covers this arch type.

ajcraig: resolved in latest commit.

proposals/sup_device_roles_to_capabilities.md
ajcraig: I'd like to table "make capabilities more verbose" for another SUP. The community has gone into rabbit holes on how verbose / and what vocabulary to utilize for this. My opinion, we should let a follow up SUP tackle that task.

proposals/sup_device_roles_to_capabilities.md
ajcraig: ditto on the comment from above.

proposals/sup_device_roles_to_capabilities.md
ajcraig: ditto on comment above.

proposals/sup_device_roles_to_capabilities.md
singhmj-1: sounds good!

proposals/sup_device_roles_to_capabilities.md
singhmj-1: sure!

proposals/sup_device_roles_to_capabilities.md
```diff
@@ -0,0 +1,221 @@
...
+| interfaces               | array             | N         | Communication interfaces present on the device. Each entry includes `type` (see [CommunicationInterfaceType](#communicationinterfacetype)). |
+| otelCollector            | boolean           | Y         | Reports whether an OTEL collector is present on the device. |
+| supportedRuntimes        | array             | Y         | Standard Margo OCI runtimes available on the device. See [SupportedRuntimes](#supportedruntimes). A device MUST report at least one entry. |
+| supportedDeploymentTypes | array             | Y         | Manifest/deployment formats the device can receive and process locally. See [supportedDeploymentTypes](#supporteddeploymenttypes). A device MUST report at least one entry. |
+
+## Enumerations
+
+> Note: These enumerations can be expanded in the future through the SUP process. This SUP aims to set the framework for future expansion. 
+
+### CpuArchitectureType
```
srberard, line 125: Minor nit:  It may be worth aligning with what OCI does here (as this is what Compose and k8s already do).  They define the architecture here:  https://github.com/opencontainers/image-spec/blob/main/config.md#properties which is really a reference to Golang's definition here:  https://go.dev/doc/install/source#environment

The values you have below are the same/consistent so this is just a refinement.  If we want to support a subset of that it's fine, but I would still refer back to the OCI language as the official list to select from.

phil-abb:
Merging in this SUP since it was approved: https://github.com/margo/specification-enhancements/issues/63#issuecomment-4673479535

PR #54, opened 2026-04-20
SUP: Device specific parameter values

arne-broering:
Addressing issue: https://github.com/margo/specification/issues/141.

Today, the Margo **Application Description** explicitly models parameters as values that the end user provides when installing/updating the application. A mechanism for **device-supplied parameter values** is currently missing. Without such a mechanism, the user has to discover these values manually and pass them as parameters, even though the device may already know them. 

The proposal is to add a new optional `valueFrom` field to `Parameter` in the  **Application Description** and instantiate the parameter value from what is defined under the new element `installContext` in the `DeviceCapabilitiesManifest`. Today `Parameter` has `name`, `value` (default value), and `targets`. The proposal is to add a `valueFrom` field that indicates the value is resolved from another source. 
The `valueFrom` mechanism is generic and can be in future applied to other sources, not only devices.
If both, `value` and `valueFrom`, are defined, `value` acts as the fallback and will be used if resolution fails.
A new element called `installContext` is added to the `DeviceCapabilitiesManifest`  and lists parameters supplied by the device.

### DIFF
--- a/proposals/device-specific-parameter-values.md
+++ b/proposals/device-specific-parameter-values.md
@@ -72,7 +72,6 @@ valueFrom:
   device:
     key: margo.cluster.hostname
     required: true
-    fallback: null
 ```
 
 **Proposed semantics**
@@ -124,7 +123,7 @@ Before the device installs or updates an application, the device’s Workload Fl
 
 The resolved value MUST satisfy the schema referenced by the configuration setting exactly as if the value had been provided by the user. This keeps the current validation model unchanged. 
 
-If a referenced device key cannot be resolved: if a `fallback` is defined, use it; otherwise installation/update MUST fail with a clear validation/resolution error.
+If a referenced device key cannot be resolved, then the `value` is used as a fallback, if it is defined, otherwise installation/update MUST fail with a clear validation/resolution error.
 
 
 ```json
--- a/proposals/device-specific-parameter-values.md
+++ b/proposals/device-specific-parameter-values.md
@@ -151,10 +151,19 @@ If a referenced device key cannot be resolved, then the `value` is used as a fal
 ### Margo-reserved Keys
 
 * `margo.cluster.hostname` - For ingress hosts, cert DNS names, externally reachable application URLs.
-
 * `margo.cluster.defaultStorageClass` - For PersistentVolume/PersistentVolumeClaim defaults on Kubernetes devices.
 
+Could be extended by further semantically-enriched keys, such as:
 
+* `margo.network.defaultExternalHostname` - concrete full hostname (FQDN) that the platform recommends or provides as the default external hostname. E.g., hostname = `edge01.plant-a.example.com`.
+* `margo.network.ingressBaseDomain` - base domain under which application-specific hostnames are created. In case an application needs a distinct hostname by extending the base domain. E.g., base domain = `apps.plant-a.example.com` and application can derive concrete hostname = `factory-insights.apps.plant-a.example.com`.
+
+_Alternatively:_ defining environment specific keys:
+
+* `margo.k8s.storageClassName` - The Kubernetes storageClassName that an application should use for its PersistentVolumeClaims.
+* `margo.k8s.ingressClassName` - The Kubernetes ingressClassName that an application should put on its Ingress resources. 
+* `margo.k8s.namespace` - Tthe Kubernetes namespace into which the application should be deployed (e.g., "margo.k8s.namespace": "factory-apps").
+* `margo.k8s.serviceAccountName` - The Kubernetes ServiceAccount name that application pods should run as (e.g., "margo.k8s.serviceAccountName": "margo-workload").
 
 ### Example 
 
@@ -170,7 +179,7 @@ The resolved values are then injected into:
 
 * Helm values (ingress.host, certificate.dnsNames[0], persistence.storageClass)
 or 
-* Compose environment variables (APP_HOSTNAME, DEVICE_ID)
+* Compose environment variables (APP_HOSTNAME)
 
 
 
@@ -379,7 +388,7 @@ Inclusion of new `installContext` element, as a dictionary of device-supplied pa
           "type": "wifi"
         }
       ]
-    },
+    },   
     "installContext": {
       "margo.cluster.hostname": {
         "type": "string",
@@ -398,7 +407,7 @@ Inclusion of new `installContext` element, as a dictionary of device-supplied pa
         "value": "standard",
         "description": "Default Kubernetes StorageClass to use for application PVCs when the application requests persistent storage.",
         "mutable": true
-      }
+      }    
     }
   }
 }
--- a/proposals/device-specific-parameter-values.md
+++ b/proposals/device-specific-parameter-values.md
@@ -1,0 +1,455 @@
+# Specification Update Proposal
+
+## Owner
+
+> List the name(s) of the person driving the SUP to completion.
+
+Arne Broering, Siemens AG
+
+> Complete as part of Phase 2: Proposal Creation
+
+## Summary
+
+> Provide a summary (in layman's terms) explaining the changes the SUP is proposing
+> 
+> Complete as part of Phase 2: Proposal Creation
+
+Today, the Margo **Application Description** explicitly models parameters as values that the end user provides when installing/updating the application. A mechanism for **device-supplied parameter values** is currently missing. Without such a mechanism, the user has to discover these values manually and pass them as parameters, even though the device may already know them. 
+
+The proposal is to add a new optional `valueFrom` field to `Parameter` in the  **Application Description** and instantiate the parameter value from what is defined under the new element `installContext` in the `DeviceCapabilitiesManifest`. Today `Parameter` has `name`, `value` (default value), and `targets`. The proposal is to add a `valueFrom` field that indicates the value is resolved from another source. 
+The `valueFrom` mechanism is generic and can be in future applied to other sources, not only devices.
+If both, `value` and `valueFrom`, are defined, `value` acts as the fallback and will be used if resolution fails.
+A new element called `installContext` is added to the `DeviceCapabilitiesManifest`  and lists parameters supplied by the device.
+
+
+## Reason for proposal
+
+> Explain why this SUP is needed and how it improves on what we have in the specification
+> 
+> Complete as part of Phase 2: Proposal Creation
+
+This SUP is needed because the current Margo specification already recognizes application parameters as a first-class concept, but it models them almost entirely as values that are provided by the user at install/update time. Currently, configurable parameters are defined so the end user can specify values. 
+
+Today, a parameter can target Helm values using dot notation or Compose environment variables, which is good because it lets an application developer say where the value must land during deployment. But the spec does not define a standard mechanism for saying that a value should come from the device itself rather than from manual user input. That means the spec handles parameter injection **but does not handle parameter resolution from device context**.
+
+The device capabilities model already requires the device to report characteristics such as role, CPU, memory, storage, peripherals, and interfaces to the Workload Fleet Manager (WFM), and it says the device client must update this information if it changes. However, those reported capabilities are primarily for describing the device and matching workloads, not for supplying install-time values like a hostname or the default storage class name. 
+
+Examples of such device-supplied parameter values are related to its communication or computation specifics:
+* The application needs to know which **hostname** or **IP address** the device uses. Then the application is able to use it as external endpoint (ingress) or within a TLS certificate. If Margo had a device-supplied parameter for the device's hostname (or IP), the device could resolve it once and use it for deployment. That would spare the user from having to discover and enter the value manually.
+* The user deploying a Margo Helm-based application may not know which Kubernetes **storage class** (e.g., standard, fast-ssd) exists on a device, or which one is marked as the default. That is why the device can often discover the default storage class more reliably than the user.
+* The user deploying a Margo Compose-based application may not know default **host path for app data**, or the default **named volume**, and those values could be device-supplied.
+
+
+
+## Requirements alignment acknowledgment
+
+> An acknowledgment that the SUP meets minimum requirements and doesn't introduce any requirements that are out of Margo's scope or vision. This section must have link(s) to applicable features and a statement about any requirements that were agreed to be out of scope for the SUP.
+> 
+> Complete as part of Phase 2: Proposal Creation
+
+This SUP addresses the identified and agreed upon feature gap 'Define how device-specific parameter values can be provided': https://github.com/margo/specification/issues/141
+
+
+
+## Technical proposal
+
+> The SUP's technical details. There must be enough technical details that someone can take the information in this section and implement it on their own.
+> 
+> Complete as part of Phase 3: SUP Technical Development
+
+### Adding a new optional `valueFrom` field to Parameter
+
+Today Parameter has the attributes [link](https://docs.margo.org/specification/applications/application-description#defining-configurable-application-parameters):
+
+* name
+* value (default value)
+* targets (to indicate which component the value should be applied to)
+
+This proposal adds:
+
+``` yaml
+valueFrom:
+  device:
+    key: margo.cluster.hostname
+    required: true
+    fallback: null
+```
+
+**Proposed semantics**
+
+Extension of `parameters`:
+
+* `valueFrom` (optional): Indicates that the parameter value is to be resolved from a non-user source before installation or update.
+* For GA1, the only standard source is `device`.
+* `device.key` (required): Identifier of the device-supplied parameter to resolve.
+* `device.required` (optional): If true, installation MUST fail if the key cannot be resolved. Default is true.
+* The resolved `value` MUST still be validated against the parameter’s configured schema.
+* If both, `value` and `valueFrom`, are defined, `value` acts as the fallback and will be used if resolution of `valueFrom` fails.
+
+That keeps the existing parameter/configuration model intact and simply adds a second source of values.
+
+Parameters that are expected to be supplied by the device are added to the `parameters` element of the **ApplicationDescription** with the `valueFrom` indicator specified as `device` as shown below:
+
+```yaml
+kind: ApplicationDescription
+...
+parameters:
+  ingressHost:
+    name: ingressHost
+    # PROPOSED EXTENSION: tells Margo that the device supplies this value
+    valueFrom:
+      device:
+        key: margo.cluster.hostname
+        required: true
+    targets:
+      # Helm values.yaml target(s)
+      - pointer: ingress.host
+        components:
+          - fi-web-helm
+      - pointer: certificate.dnsNames[0]
+        components:
+          - fi-web-helm
+      # Compose env var target(s)
+      - pointer: APP_HOSTNAME
+        components:
+          - fi-web-compose
+```
+
+
+### Adding device supplied parameters to DeviceCapabilities
+
+The `DeviceCapabilitiesManifest` is extended by adding an element called `installContext`. This element lists the device supplied parameters.
+
+Before the device installs or updates an application, the device’s Workload Fleet Management Client resolves all parameters that use `valueFrom.device`. This is done prior to invoking the deployment provider (e.g., Helm or Compose). The device MUST resolve all `valueFrom.device` references to concrete values.
+
+The resolved value MUST satisfy the schema referenced by the configuration setting exactly as if the value had been provided by the user. This keeps the current validation model unchanged. 
+
+If a referenced device key cannot be resolved: if a `fallback` is defined, use it; otherwise installation/update MUST fail with a clear validation/resolution error.
+
+
+```json
+{
+  "apiVersion": "device.margo.org/v1alpha1",
+  "kind": "DeviceCapabilitiesManifest",
+  ...
+  "installContext": {
+    "margo.cluster.hostname": {
+      "type": "string",
+      "value": "edge01.plant-a.example.com",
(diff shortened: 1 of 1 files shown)


phil-abb:
@margo/technical-wg A new SUP has been started by @arne-broering. If anyone is interested in helping work on fleshing out the details of this SUP, please contact @arne-broering

proposals/device-specific-parameter-values.md
```diff
@@ -0,0 +1,463 @@
...
+## Summary
+
+> Provide a summary (in layman's terms) explaining the changes the SUP is proposing
+> 
+> Complete as part of Phase 2: Proposal Creation
+
+Today, the Margo **Application Description** explicitly models parameters as values that the end user provides when installing/updating the application. A mechanism for **device-supplied parameter values** is currently missing. Without such a mechanism, the user has to discover these values manually and pass them as parameters, even though the device may already know them. 
+
+The proposal is to add a new optional `valueFrom` field to `Parameter` in the  **Application Description** and instantiate the parameter value from what is defined under the new element `installContext` in the `DeviceCapabilitiesManifest`. Today `Parameter` has `name`, `value` (default value), and `targets`. The proposal is to add a `valueFrom` field that indicates the value is resolved from another source. 
+The `valueFrom` mechanism is generic and can be in future applied to other sources, not only devices.
```
phil-abb, line 20: Do you have any other sources in mind?

proposals/device-specific-parameter-values.md
```diff
@@ -0,0 +1,463 @@
...
+
+Today, a parameter can target Helm values using dot notation or Compose environment variables, which is good because it lets an application developer say where the value must land during deployment. But the spec does not define a standard mechanism for saying that a value should come from the device itself rather than from manual user input. That means the spec handles parameter injection **but does not handle parameter resolution from device context**.
+
+The device capabilities model already requires the device to report characteristics such as role, CPU, memory, storage, peripherals, and interfaces to the Workload Fleet Manager (WFM), and it says the device client must update this information if it changes. However, those reported capabilities are primarily for describing the device and matching workloads, not for supplying install-time values like a hostname or the default storage class name.
```
phil-abb: Regarding "the default stroage class name."

I think we should move away from referring to this as the default storage class because the storage class name the device wants apps to use may not be the cluster's configured default storage class.

proposals/device-specific-parameter-values.md
```diff
@@ -0,0 +1,463 @@
...
+
+Examples of such device-supplied parameter values are related to its communication or computation specifics:
+* The application needs to know which **hostname** or **IP address** the device uses. Then the application is able to use it as external endpoint (ingress) or within a TLS certificate. If Margo had a device-supplied parameter for the device's hostname (or IP), the device could resolve it once and use it for deployment. That would spare the user from having to discover and enter the value manually.
+* The user deploying a Margo Helm-based application may not know which Kubernetes **storage class** (e.g., standard, fast-ssd) exists on a device, or which one is marked as the default. That is why the device can often discover the default storage class more reliably than the user.
```
phil-abb:
```suggestion
* The user deploying a Margo Helm-based application may not know which Kubernetes **storage class** (e.g., standard, fast-ssd) exists on a device, or which one the device vendor wants apps to use. That is why the device can often discover the storage class more reliably than the user.
```

proposals/device-specific-parameter-values.md
```diff
@@ -0,0 +1,463 @@
...
+* The user deploying a Margo Compose-based application may not know default **host path for app data**, or the default **named volume**, and those values could be device-supplied.
```
phil-abb, line 40: same here, maybe we don't use "default"

proposals/device-specific-parameter-values.md
```diff
@@ -0,0 +1,463 @@
...
+        components:
+          - fi-web-compose
+```
+
+
+### Adding device supplied parameters to DeviceCapabilities
+
+The `DeviceCapabilitiesManifest` is extended by adding an element called `installContext`. This element lists the device supplied parameters.
+
+Before the device installs or updates an application, the device’s Workload Fleet Management Client resolves all parameters that use `valueFrom.device`. This is done prior to invoking the deployment provider (e.g., Helm or Compose). The device MUST resolve all `valueFrom.device` references to concrete values.
```
phil-abb: I have some questions/concerns about this. Why does the WFM have to resolve the values? By saying this, we're requiring the device to send the values to the WFM. I'm not convinced sending the value to the WFM is a good thing, especially when we talk about values that contain sensitive/secret information. This would require some form of encryption to send these values to the WFM, and it introduces an additional attack vector. 

Even taking into account the [SUP to move template rendering to the WFM](https://github.com/margo/specification-enhancements/pull/40) I think it's better if the device is responsible for providing the value and not the WFM that way the values don't leave the device.

proposals/device-specific-parameter-values.md
```diff
@@ -0,0 +1,463 @@
...
+
+The resolved value MUST satisfy the schema referenced by the configuration setting exactly as if the value had been provided by the user. This keeps the current validation model unchanged.
```
phil-abb: The schema is currently only applied to the configuration section, which only applies to the parameters expecting input from users. If there is an intention to allow for a schema, then there needs to be additional changes. 

Also, if the intention is to use the schema, then we wouldn't need the `required` property under `valueFrom.device` since the schema can be used to enforce this rule.

proposals/device-specific-parameter-values.md
```diff
@@ -0,0 +1,463 @@
...
+The `DeviceCapabilitiesManifest` is extended by adding an element called `installContext`. This element lists the device supplied parameters.
+
+Before the device installs or updates an application, the device’s Workload Fleet Management Client resolves all parameters that use `valueFrom.device`. This is done prior to invoking the deployment provider (e.g., Helm or Compose). The device MUST resolve all `valueFrom.device` references to concrete values.
+
+The resolved value MUST satisfy the schema referenced by the configuration setting exactly as if the value had been provided by the user. This keeps the current validation model unchanged. 
+
+If a referenced device key cannot be resolved, then the `value` is used as a fallback, if it is defined, otherwise installation/update MUST fail with a clear validation/resolution error.
+
+
+```json
```
phil-abb, line 126: Apart from my comment above about sending the value, I question the need to send this info to the WFM. These parameters should be transparent to the user because they shouldn't be changed, and the user doesn't need to know what they are. 

The only value I see here is so that a pre-check could be done to see if the device has the expected required parameter, or not. In order to do this check, the other information isn't needed.

If we define the required parameters, then the parameter should be available. The situation where they wouldn't be available is if a device is making custom parameters available.  In this case, it would only work with applications that know the device is making them available anyway.

proposals/device-specific-parameter-values.md
```diff
@@ -0,0 +1,463 @@
...
+      "mutable": true
+    },
+  ...
+```
+
+
+### Margo-reserved Keys
+
+* `margo.cluster.hostname` - For ingress hosts, cert DNS names, externally reachable application URLs.
+* `margo.cluster.defaultStorageClass` - For PersistentVolume/PersistentVolumeClaim defaults on Kubernetes devices.
```
phil-abb:
```suggestion
* `margo.cluster.storageClass` - For PersistentVolume/PersistentVolumeClaim defaults on Kubernetes devices.
```

proposals/device-specific-parameter-values.md
```diff
@@ -0,0 +1,463 @@
...
+      "description": "Default ingress class to use for applications that expose HTTP/HTTPS endpoints through Kubernetes ingress.",
+      "mutable": true
+    },
+  ...
+```
+
+
+### Margo-reserved Keys
+
+* `margo.cluster.hostname` - For ingress hosts, cert DNS names, externally reachable application URLs.
```
phil-abb: Do we need the "margo" prefixes? What is the reasoning behind adding this prefix?

proposals/device-specific-parameter-values.md
```diff
@@ -0,0 +1,463 @@
...
+* `margo.cluster.defaultStorageClass` - For PersistentVolume/PersistentVolumeClaim defaults on Kubernetes devices.
+
+Could be extended by further semantically-enriched keys, such as:
+
+* `margo.network.defaultExternalHostname` - concrete full hostname (FQDN) that the platform recommends or provides as the default external hostname. E.g., hostname = `edge01.plant-a.example.com`.
```
phil-abb: What is the difference between `margo.cluster.hostname` and `margo.network.defaultExternalHostname`?

proposals/device-specific-parameter-values.md
```diff
@@ -0,0 +1,463 @@
...
+
+### Example 
+
+In the `ApplicationDescription` and `DeviceCapabilitiesManifest` examples below, the WFM has through the `ApplicationDescription` the parameters that use `valueFrom.device.key`. The current specification already defines the overall deployment model in which applications are described via `ApplicationDescription` and then deployed with chosen parameters.
+The device has already reported its capabilities, now including the proposed `installContext` dictionary. The current spec already has the device capability reporting flow.
+
+Before installation, the device-side WFMC resolves:
+
+* `ingressHost` from `margo.cluster.hostname`
+* `storageClass` from `margo.cluster.defaultStorageClass`
```
phil-abb:
```suggestion
* `storageClass` from `margo.cluster.storageClass`
```

proposals/device-specific-parameter-values.md
```diff
@@ -0,0 +1,463 @@
...
+Parameters that are expected to be supplied by the device are added to the `parameters` element of the **ApplicationDescription** with the `valueFrom` indicator specified as `device` as shown below:
+
+```yaml
+kind: ApplicationDescription
+...
+parameters:
+  ingressHost:
+    name: ingressHost
+    # PROPOSED EXTENSION: tells Margo that the device supplies this value
+    valueFrom:
```
phil-abb, line 98: Does this need to be on the "target" instead of at the parameter root? Since it's possible to have multiple deployment profiles (e.g., one for Compose and one for Kubernetes), you will potentially have targets for Compose and targets for Kubernetes, and each one might need its own valueFrom that it maps to the correct key. If the keys are different between Kuberentes and Compose, or whatever else we add later, it becomes a bit of a problem.

I don't know, I guess it depends on the keys we define and how common they are between the different runtimes. I guess you could also use different parameters if the keys are different.

phil-abb:
An alternative that was discussed on the side was around potentially using a simplified expression language (something similar to what Helm has, but more basic) instead of adding the `valueFrom` object. 
 
For example:
- ${default "localhost" margo.cluster.hostname}
- ${required "Error: hostname must be available" margo.cluster.hostname}
 
We could keep the functions very basic to start with and add more as more use cases are known.
 
It does add additional complexity, but at the same time, it adds more flexibility for adding advanced logic such as greater-than, less-than, and ternary operations.

proposals/device-specific-parameter-values.md
arne-broering: Another source could be the WFM providing information about the deployment context (the values are hold by a WFM-managed deployment context object). Potentially e.g. also from a secret manager or other components in the ecosystem.

proposals/device-specific-parameter-values.md
arne-broering: yes, that is right. we want to specify what storage class should this application use on this device. So I will adjust this (and the examples) to something like "device-preferred storage class".

proposals/device-specific-parameter-values.md
arne-broering: You are right, if the device can resolve a value locally, and the WFM does not need the literal value for scheduling or UX purposes, then giving that value to the WFM is unnecessary privilege escalation. So I agree, it is better to propose a "device-local resolution" as far as possible. I will adjust this.

proposals/device-specific-parameter-values.md
phil-abb: I do have a use case for the value coming from the WFM.

What I need is for either the device or the WFM to assign a random ID (UUID) that is never changed once it's set (immutable). Since we cannot use the "lookup" function, during the initial deployment, the ID needs to be generated and reused from that point forward. Initially, I was thinking this is something the device could do, but for something like this, it might make more sense for the WFM to do it.

phil-abb:
I can think of some other keys, but it related to having specific operators installed, so I think it would be better to wait until we figure out how to handle operators in an interoperable way before I suggest them

jjaswanson4:
This is a good idea, allowing for per-device rendering of specific values makes application deployment much more scalable.

Is the intention to connect this with information coming from what the devices are reporting about themselves to the WFM?

For example, if the device reports `hostname: device1`, then this renders accordingly.

> Even taking into account the SUP to move template rendering to the WFM I think it's better if the device is responsible for providing the value and not the WFM that way the values don't leave the device.

We're already proposing/requiring a bunch of information be sent from the devices up to the WFM, including information about the device, capabilities, hardware, etc etc - Why not do the rendering where compute power is plentiful?

proposals/device-specific-parameter-values.md
nseba: I think it's a good idea to have the margo prefix. Actually also with the device capabilities I would go for the same approach. The reason is twofold: 

1. It makes it clear which keys are margo (general target) and reserves them so that they are not misused
2. It clearly identifies vendor/app/device-specific values which make sense in that specific context without the risk of overlapping between different implementations (same name, different meanings).

phil-abb:
> We're already proposing/requiring a bunch of information be sent from the devices up to the WFM, including information about the device, capabilities, hardware, etc etc - Why not do the rendering where compute power is plentiful?

@jjaswanson4 

My main concern here is for sensitive information. If we're forcing information to be sent from the device to the WFM to do the value replacement in the WFM, it means that at some point, we'll need to send sensitive information. This means we're increasing the attack surface. We would need to ensure we are securing the information and provide a way to indicate which information is secured and how it is handled. If we don't force the device to send this data, then it's not a problem we have to solve. It's more of a question of risk vs. reward. 

You could make the argument that we may never need to send sensitive data, but if there is an intention for device vendors and app vendors to use their own keys (via collaboration) and not only Margo-defined keys, then the need for being able to handle sensitive information increases.

jjaswanson4:
> My main concern here is for sensitive information. If we're forcing information to be sent from the device to the WFM to do the value replacement in the WFM, it means that at some point, we'll need to send sensitive information. This means we're increasing the attack surface. We would need to ensure we are securing the information and provide a way to indicate which information is secured and how it is handled. If we don't force the device to send this data, then it's not a problem we have to solve. It's more of a question of risk vs. reward.

> You could make the argument that we may never need to send sensitive data, but if there is an intention for device vendors and app vendors to use their own keys (via collaboration) and not only Margo-defined keys, then the need for being able to handle sensitive information increases.

This is, IMO, the one good argument for what data to _not_ send to a WFM and keep local to the device, and to do something like this that's rendered _only_ on the device.

I'm fine with that, with the caveat of if we do allow for that, we should use simple text/replace or something very lightweight to handle it, and still have the WFM do most of the "heavy lifting".

However you could argue we should be introducing specific controls around devices and their information within the WFM - everything from RBAC to DB encryption to one-way encrypted fields, etc etc - there are ways to handle sensitive information.

I think we're going to end up having to discuss the latter anyway, as different organizations have different requirements around what they consider "sensitive" - I've seen organizations require IP addresses be stored as sensitive information, similar to passwords/tokens/etc - and we should contemplate giving end users that level of control, should they chose to use it.

phil-abb:
> I'm fine with that, with the caveat of if we do allow for that, we should use simple text/replace or something very lightweight to handle it, and still have the WFM do most of the "heavy lifting".

Yeah, I think that makes sense

> However you could argue we should be introducing specific controls around devices and their information within the WFM - everything from RBAC to DB encryption to one-way encrypted fields, etc etc - there are ways to handle sensitive information.
> 
> I think we're going to end up having to discuss the latter anyway, as different organizations have different requirements around what they consider "sensitive" - I've seen organizations require IP addresses be stored as sensitive information, similar to passwords/tokens/etc - and we should contemplate giving end users that level of control, should they chose to use it.

Yeah, there are things we still need to figure out with this

proposals/device-specific-parameter-values.md
arne-broering: That is a good point: right now 'configuration' binds a `parameter` to a `schema`. The `configuration` section makes sense for user defined parameter values, as it describes the structure of a UI. For device defined parameters, it would not be necessary to display it in a UI, but could be helpful to get them confirmed by a human. So, just to allow the connection to a schema, we should not include the device specified parameters in the `configuration` section. We could include a `schema` link already in the `parameter` definition. What was your reasoning to connect a `parameter` with a `schema` only at the `configuration` level? Would you be open to moving the link to the `schema` into the `parameter`? It would be strange to do that only for device-defined parameters.

phil-abb: The reason the schema was bound to the configuration section is that that is where the customer provides input. App vendors would know what to put as default values so that they don't need to be validated. Customers may not know exactly what to put, so they could accidentally provide an invalid value.

I'm not sure if we'd need to use the schema for these new device-specific parameter values. If we're defining what the expressions are, then to a large degree, we are setting the expectation for what the value should be.

arne-broering: ok, I took this out.

proposals/device-specific-parameter-values.md
arne-broering: (_Just to be clear, the 'value' mentioned in the quoted sentence is supposed to refer to the 'value' under 'parameter' in the Application Description._)

I guess you are correct, if the device always informs about the parameters, which Margo defines as "required" if available for a device. My feeling is that this would probably require a mechanism that checks / enforces the provision of all available & required parameters automatically. Otherwise device owners may forget to inform about them.

I just thought of a different situation, as an extension of the example [file here](https://github.com/margo/specification/blob/pre-draft/src/specification/applications/resources/examples/valid/ApplicationDescription-002.yaml).  The ApplicationDescription has a parameter 'cpuLimit' ("Maximum number of CPU cores to allow the application to consume") that is by default set to 2. Now, this SUP's proposed extension would allow the device to have a say as well: let's say we define a key like 'margo.device.cpuLimit' that would allow the device to define the cpuLimit is 3 (specified in the DeviceCapabilities). So looking back at the sentence quoted by you above, the WFMC would first take it from the device and would fallback on the default 'value' if a device does not specify 'margo.device.cpuLimit'.

```yaml
parameters:
  cpuLimit:
    value: 2
    targets:
      - pointer: settings.limits.cpu
        components: ["digitron-orchestrator"]
    valueFrom:
      device:
        key: margo.device.cpuLimit
        required: true
```

Similar situations could be thought of with the already discussed keys for (e.g., storage class).

proposals/device-specific-parameter-values.md
arne-broering: Yes, the details of the keys still need to be straightened out and their usage needs to be elaborated.
Maybe some need to be dropped. And we should think of further parameter keys that are needed.

phil-abb: Yeah, maybe we should keep the keys in the original SUP limited to ones people have specific use cases for, and then as new use cases come up, more can be added later. I don't think we need to define all the ones we'll have with this initial SUP. This SUP can focus more on setting the pattern for how we do it.

proposals/device-specific-parameter-values.md
phil-abb: >I guess you are correct, if the device always informs about the parameters, which Margo defines as "required" if available for a device. My feeling is that this would probably require a mechanism that checks / enforces the provision of all available & required parameters automatically. Otherwise device owners may forget to inform about them.

This could be incorporated into the conformance test suite. For the parameter keys we define, we can set expectations and validate in our conformance suite that the key is present and the value matches an expected pattern/type.

> I just thought of a different situation, as an extension of the example [file here](https://github.com/margo/specification/blob/pre-draft/src/specification/applications/resources/examples/valid/ApplicationDescription-002.yaml)...

This is where [making things expression-based](https://github.com/margo/specification-enhancements/pull/54#issuecomment-4282861112) has its advantages. I'm not sure about your specific use case example, but I'm sure there will be similar cases where someone may want it to take the greater or lesser of two values. One provided by the customer and one provided by the device. Or, be able to fall back to a user-provided value if the device doesn't have it.

proposals/device-specific-parameter-values.md
arne-broering: Interesting idea, and you are right, we need to be able to define parameter values specific per deployment profile. However, I just tried it out to move the `valueFrom` branch on the target and it ended up in a quite verbose style of defining things. My preference would be to define the parameters already per deployment profile. We would also have to do that with the current spec approach of user defined values.

proposals/device-specific-parameter-values.md
arne-broering: This is also an interesting use case. Nonetheless, as an initial version for GA1, I suggest to keep it simple and only support "device" as a source.

proposals/device-specific-parameter-values.md
arne-broering: I justed tidied up the margo reserved keys. Maybe ou can take a look again to see if that covers everything that we want to support for GA1.

phil-abb: `margo.cluster.hostname` and `margo.cluster.storageClass` are the keys I currently have use cases for, so it covers what I know about.

phil-abb:
@margo/technical-wg 

This device-specific parameter value, SUP, is in final review. Please review the proposal and add any questions/comments to the PR.

phil-abb:
This SUP was approved, so I'm force-merging it in: https://github.com/margo/specification-enhancements/issues/87#issuecomment-5319886648

PR #55, opened 2026-04-22
SUP: Device specific runtimes

arne-broering:
This SUP addresses: https://github.com/margo/specification/issues/93

This makes Margo more practical for real-world industrial ecosystems where some apps are not fully generic, but depend on a vendor-specific software environment or runtime functionality.
Without this feature, an app may look deployable because the device has enough CPU/memory and the right general role, even though it is still missing the vendor runtime the app actually needs. 
With this new feature, the app developer can describe these requirements in a standard, machine-readable way, and the WFM can enforce them automatically.


phil-abb:
@margo/technical-wg A new SUP has been started by @arne-broering. If anyone is interested in helping work on fleshing out the details of this SUP, please contact @arne-broering

phil-abb:
@ajcraig Is this SUP addressing what is meant by [this issue](https://github.com/margo/specification/issues/93)? This SUP addresses supporting custom operating system requirements, but I had the impression that the issue was about supporting custom deployment options (alternatives to Helm/Compose). After reading this SUP and reading through the issue, it's no longer clear to me what the intention of the issue was.

cc @arne-broering

arne-broering:
> @ajcraig Is this SUP addressing what is meant by [this issue](https://github.com/margo/specification/issues/93)? This SUP addresses supporting custom operating system requirements, but I had the impression that the issue was about supporting custom deployment options (alternatives to Helm/Compose). After reading this SUP and reading through the issue, it's no longer clear to me what the intention of the issue was.
> 
> cc @arne-broering

@phil-abb thanks for your comment. yes, indeed, custom deployment options (alternatives to Helm/Compose) are not covered. 
Instead, vendor specific runtimes should be supported, which run the app still in Helm / Compose, but have some capabilities that should be considered before the WFM can allow to install an app on that device.
For example (see also yaml [here](https://github.com/margo/specification-enhancements/blob/device-specific-runtime/proposals/device-specific-runtime.md)), an app requires a specific device version, or vendor OS version (which runs Compose), or other installed apps (e.g., a vision AI may require an installed camera connector).

@ajcraig are we sure about this feature request of alternative deployment options and what examples come to your mind?

When looking e.g. at these points in the current spec:
- https://docs.margo.org/specification/applications/application-description#componentproperties-attributes
- https://docs.margo.org/specification/applications/application-description#target-attributes
it would appear to me very difficult to define a generic approach that allows _any_ alternative (to Helm/Compose). The current spec has concrete dependencies to these two standards. I.e., if _one concrete_ alternative manifest option should be supported, we would have to work that into the spec in various places. 
Or am i getting this wrong?

ajcraig:
> @phil-abb thanks for your comment. yes, indeed, custom deployment options (alternatives to Helm/Compose) are not covered.
Instead, vendor specific runtimes should be supported, which run the app still in Helm / Compose, but have some capabilities that should be considered before the WFM can allow to install an app on that device.
For example (see also yaml [here](https://github.com/margo/specification-enhancements/blob/device-specific-runtime/proposals/device-specific-runtime.md)), an app requires a specific device version, or vendor OS version (which runs Compose), or other installed apps (e.g., a vision AI may require an installed camera connector). @ajcraig are we sure about this feature request of alternative deployment options and what examples come to your mind?

Limiting this sup to custom runtime "Capability" outlined in your proposal makes sense to me. 

I am working on the new roles to capabilities sup and am consciously separating `supportedRuntimes` and `supportedManifests`. 

I see this Sup as expanding the `supportedRuntimes` by introducing `vendorRuntimes` information. 
See example [here](https://github.com/margo/specification-enhancements/blob/sup_roles_to_capabilities/proposals/sup_device_roles_to_capabilities.md#capabilities-attributes)

IMO, this aligns with what the child epic describes:

> I propose that instead, the Device Manifest declare a set of runtime environments it supports. The standard would define a standard set of runtime environments (e.g. Docker Compose and Helm), but the Device Vendor could include new or custom runtime environments as well (e.g. WASM or VendorSpecificRuntime).
An app vendor could then define what runtime their app requires, be it one of the standard ones or a new one or a vendor specific runtime.

Lastly I think this is customruntimes, not necessarily custom "OS" since the operating system isn't today something a workload can be deployed to. It's the `supportedRuntimeType`. 


@arne-broering @phil-abb thoughts?

phil-abb:
@arne-broering / @ajcraig - If the goal here is to allow an app vendor to work with a device vendor to deploy an app to something "unique", could we simplify this by having a "labels" or "tags" dictionary of key-value pairs (allowing values of string, numbers, booleans, and arrays of those types)? 

Do we need to care how they want to match an app to a device?

Is there value in us defining the specific parameters needed? There is a risk of needing to keep adding new things if we do define them.

ajcraig:
> @arne-broering / @ajcraig - If the goal here is to allow an app vendor to work with a device vendor to deploy an app to something "unique", could we simplify this by having a "labels" or "tags" dictionary of key-value pairs (allowing values of string, numbers, booleans, and arrays of those types)?
> 
> Do we need to care how they want to match an app to a device?
> 
> Is there value in us defining the specific parameters needed? There is a risk of needing to keep adding new things if we do define them.

I agree with your sentiment here. When I was creating multiple TWG features I saw some overlap between these two:
- https://github.com/margo/specification/issues/144
- https://github.com/margo/specification/issues/93

I think the key-value pair strategy could enable "any" extensions that enable the WFM to match apps to devices. Could be runtimes, could be something else. 

@arne-broering  thoughts? we could merge the two TWG features. I have some hesitation with the proposal regarding the verbosity of the customRuntime compared to our standard `supportedRuntimes` detailed in my SUP [here](https://github.com/margo/specification-enhancements/blob/sup_roles_to_capabilities/proposals/sup_device_roles_to_capabilities.md).

jjaswanson4:
IMO this more outlines a need for a `requiresFunctionality` (or equivalent) field, which results in explicit steering when deploying an application.

> I think the key-value pair strategy could enable "any" extensions that enable the WFM to match apps to devices. Could be runtimes, could be something else.
True, but I wonder if this is trying to create a specific "requires". "matchTheseLabels" seems less stringent.

I think it's worth considering how k8s does this. Quick example:
```
spec:
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: topology.kubernetes.io/zone
            operator: In
            values:
            - antarctica-east1
            - antarctica-west1
      preferredDuringSchedulingIgnoredDuringExecution:
      - weight: 1
        preference:
          matchExpressions:
          - key: another-node-label-key
            operator: In
            values:
            - another-node-label-value
```
(Not saying we should use these exact key:values, just using this for conversation)
(Since we're not really a reconciliation loop at the WFM level, I'd say ignore the "IgnoredDuringExecution" part for this discussion)

So this SUP essentially wants us to create a "requiredDuringSchedulingIgnoredDuringExecution" - basically: this app must run this device, or only on this device with this tag/label.

phil-abb:
> I agree with your sentiment here. When I was creating multiple TWG features I saw some overlap between these two:
> 
> * [Define how vendor specific extensions can be supported within the application description file + device capabilities specification#144](https://github.com/margo/specification/issues/144)
> * [Develop Margo standard feature enabling application developers to target device specific runtime environments specification#93](https://github.com/margo/specification/issues/93)
> 
> I think the key-value pair strategy could enable "any" extensions that enable the WFM to match apps to devices. Could be runtimes, could be something else.
> 
> @arne-broering thoughts? we could merge the two TWG features. I have some hesitation with the proposal regarding the verbosity of the customRuntime compared to our standard `supportedRuntimes` detailed in my SUP [here](https://github.com/margo/specification-enhancements/blob/sup_roles_to_capabilities/proposals/sup_device_roles_to_capabilities.md).

There is a fundamental difference between these two issues:

For https://github.com/margo/specification/issues/93 this is about matching, and we can enable matching in a few different ways with Margo's schema for the application definition and device capabilities documents.

https://github.com/margo/specification/issues/144 is fundamentally different because it's about allowing vendors to work independently of the Margo specification to add functionality to the application description by allowing people to add additional YAML to this document that the Margo spec knows nothing about. Of course, this will only work if the vendors are working together to know what is being added and can implement it.

phil-abb:
> I think it's worth considering how k8s does this. Quick example:

I kind of like this idea because it adds a bit more flexibility to the app vendor for what gets matched and how. We could still keep the device capabilities as simple key-value pairs.

proposals/device-specific-runtime.md
```diff
@@ -0,0 +1,296 @@
...
+  - runtimeRef: string        # required
+    vendor: string            # optional
+    vendorId: string          # optional
+    modelNumberId: string     # optional
+    version: string           # optional
+    runtimeCapabilities:      # optional
+      - string
+    properties:               # optional
+      <string>: <string | number | boolean | array>
+```
```
ajcraig:
```suggestion
```json
[
  {
    "runtimeRef": "string", // required
    "vendor": "string", // optional
    "vendorId": "string", // optional
    "modelNumberId": "string", // optional
    "version": "string", // optional
    "runtimeCapabilities": ["string"], // optional
    "properties": {
      "<string>": "<string | number | boolean | array>" // optional
    }
  }
]
```
```

File: proposals/device-specific-runtime.md [hunk 2]
```diff
@@ -0,0 +1,296 @@
...
+      capabilities:
+        - com.northstar-ida.industry.app.camera-connector # app to enable connectivity to a connected camera 
+        - com.northstar-ida.industry.app.opcua-connetor # app to access OPC UA servers 
+      properties:
+        iedkVersion: 1.24.2-1
+        dockerVersion: 20.10.24
+        operatingSystem: "Debian 12 (Bookworm)"
+        nvidiaDriverVersion: "525.85.12-1"
+        cudaDriverVersion: "12.0"
+```
```
ajcraig:
```suggestion
```json
{
  "properties": {
    "id": "ie-device-bx59a-001",
    "vendor": "Northstar-IDA Inc.",
    "modelNumber": "EdgeServer-A",
    "serialNumber": "ABC-EXAMPLE-001",
    "roles": ["standalone device"],
    "resources": {
      "cpu": [
        {
          "cores": 24,
          "architecture": "x86_64"
        }
      ],
      "memory": "30Gi",
      "storage": "900Gi",
      "peripherals": [
        {
          "type": "gpu",
          "manufacturer": "NVIDIA",
          "model": "L4"
        }
      ],
      "interfaces": [
        {
          "type": "ethernet"
        }
      ]
    },
    "runtimeEnvironments": [
      {
        "id": "com.northstar-ida.industry.edge-os",
        "vendor": "Northstar-IDA Inc.",
        "vendorId": "com.northstar-ida",
        "modelNumberId": "com.northstar-ida.edge-devices.EdgeServer-A",
        "version": "3.1.1-3-x86-64",
        "capabilities": [
          "com.northstar-ida.industry.app.camera-connector",
          "com.northstar-ida.industry.app.opcua-connector"
        ],
        "properties": {
          "iedkVersion": "1.24.2-1",
          "dockerVersion": "20.10.24",
          "operatingSystem": "Debian 12 (Bookworm)",
          "nvidiaDriverVersion": "525.85.12-1",
          "cudaDriverVersion": "12.0"
        }
      }
    ]
  }
}
```
```

**Review by ajcraig — CHANGES_REQUESTED** [review]

**Comment by ajcraig** [comment]
one thing I think would be good to collaborate on in this stage is the current [device roles to capabilities](https://github.com/margo/specification-enhancements/blob/a8cc1e28ab4c6451e71b76562876cad24c550dbb/proposals/sup_device_roles_to_capabilities.md) sup. I'm splitting out the roles into `supportedRuntimes` and `supportedManifests`. 

Example below from my SUP. 
```json
        },
        "capabilities": {
            "wfmClient": true,
            "otelCollector": true,
            "supportedRuntimes": [
                "kubernetes",
                "podman"
            ],
            "supportedManifests": [
                "helm",
                "compose"
            ]
        }
```

Follow up discussion questions:
- Just want to match our two when we propose them. I lead with `supported` and you lead with `runtime`. This is a small change discussion point. 
- I'd assume these `runtimeEnvironments` are separate from what I am proposing with `supportedRuntimes`?
- Does the custom runtime supersede or work in tandem with the traditional `supportedRuntimes` i.e. docker/kubernetes? 
    - what I mean by this question is, we must require a "application runtime" which is today OCI. And would this device specific runtime be alongside the generic app runtimes?
    - In your examples you kept the standard `roles` element. 

@phil-abb I think this is where I would like help with PMG. because this looks awful similar to [vendor extensions](https://github.com/margo/specification-enhancements/pull/56/changes). app <-->device. 
- I thought this one would be, I have an app that requires a specific application runtime that is not common. aka not OCI runtime. 
    - I'm worried this SUP is just a more verbose / different approach to your vendor extensions sup. 

(although I might be just going flat out crazy....please advise).

proposals/device-specific-runtime.md
ajcraig: just a formatting issue as device capabilities are JSON format.

proposals/device-specific-runtime.md
ajcraig: just a formatting issue as device capabilities are JSON format.

phil-abb:
> @phil-abb I think this is where I would like help with PMG. because this looks awful similar to [vendor extensions](https://github.com/margo/specification-enhancements/pull/56/changes). app <-->device.
> 
> * I thought this one would be, I have an app that requires a specific application runtime that is not common. aka not OCI runtime.
>   
>   * I'm worried this SUP is just a more verbose / different approach to your vendor extensions sup.
> 
> (although I might be just going flat out crazy....please advise).

The way I see it is that vendor extensions allow suppliers to collaborate to add additional functionality by using vendor extensions to enhance what the application definition, application deployment, and device capabilities allow for.

I see this SUP as a way to do matching between what an app wants and what a device has, but I see it in more general terms than its current form, as my comments above suggest.

samuel-greising:
Pushing back on this SUP in its current shape. My concern is that it institutionalises vendor lock-in at the schema layer, which runs against Margo's stated purpose as a vendor-neutral interoperability layer.

The lock-in concentrates in three fields:

- runtimeRef keyed to a vendor-namespaced ID (com.<vendor>.<runtime>) with no neutral equivalent
- deviceSelector.vendorId: schema-level "must be this vendor"
- deviceSelector.modelNumberIds[]: SKU-level gating

An ApplicationDescription that targets these is, by construction, non-portable across Margo-compliant fleets. If multiple vendors adopt the pattern (and they will), the practical outcome is Margo certifying devices that can't run each other's apps, i.e. the pre-Margo status quo with a logo on it. The repeated reliance on a "shared Margo/vendor runtime dictionary" that the SUP never actually defines sharpens this: whoever ends up curating that dictionary effectively controls who is a first-class runtime.

I'd argue the right move is to split this into two pieces, both of which already have homes in the SUP queue.

1. Vendor-specific advertisement belongs in @phil-abb's vendor-extensions SUP, not the core schema.

[sup_enable_vendor_extensions.md](https://github.com/margo/specification-enhancements/blob/main/proposals/sup_enable_vendor_extensions.md) already provides a principled channel for exactly the content this SUP wants to add. Under x-siemens-extensions: (or whichever namespace), a vendor can advertise IEDK presence, capability IDs, model selectors, anything they want. The load-bearing property of that mechanism is that unknown extensions MUST be ignored by uncommitted WFMs and MUST NOT break conformance, so vendor metadata can travel with the application without becoming a deployment gate against the rest of the ecosystem.
This is also the line @phil-abb himself drew on April 30: issue #93 (matching) and issue #144 (vendor extensions) are different problems. As written, this SUP is solving the matching problem using the toolkit that should be reserved for extensions, which is what causes the lock-in.

2. The legitimate matching need (capabilities, runtime versions) should be conformance-test-based, not vendor-identifier-based.

The underlying problem #93 raises is real: apps need to express that they require certain capabilities (camera I/O, OPC-UA, GPU drivers) and certain minimum runtime versions, and the WFM needs to refuse deployment where those aren't present. Vendor identifiers are the wrong primitive for that. Conformance suites are the right one:

- Margo defines a conformance suite per capability (margo.capability.camera-io, margo.capability.opcua, etc.). Any device that passes the suite advertises the label. The WFM matches on labels, never on vendor.
- Runtime version compatibility is expressed as Margo conformance levels (margo.runtime.compose >= 2), each backed by a published, re-runnable test suite. This also resolves the currently-undefined comparison between ">=3.1.0 <4.0.0" and a device-reported 3.1.1-3-x86-64.
- @ajcraig's roles-to-capabilities SUP already owns the runtime-type axis (supportedRuntimes, supportedManifests). A capability vocabulary would extend that pattern to fine-grained features.

This composes cleanly with the rest of the thread: @jjaswanson4's k8s-affinity intuition (label selectors), @phil-abb's earlier labels suggestion, and @ajcraig's runtime/manifest split. Nothing here is in tension with what's already being proposed.

Concrete suggestion

- Rework this SUP so the IEDK / EdgeServer content lives as a worked example under the vendor-extensions SUP, not in the core schema.
- Open a companion SUP for a Margo-governed capability vocabulary plus conformance-suite governance, addressing the matching half of #93 vendor-neutrally.
- Keep core-schema fields strictly to advertising capabilities (if anything), never to gating on vendor identity.

Happy to help draft either follow-up.

proposals/device-specific-runtime.md
arne-broering: ok thanks, i updated it

proposals/device-specific-runtime.md
arne-broering: thanks, I adjusted it.

arne-broering:
> @arne-broering / @ajcraig - If the goal here is to allow an app vendor to work with a device vendor to deploy an app to something "unique", could we simplify this by having a "labels" or "tags" dictionary of key-value pairs (allowing values of string, numbers, booleans, and arrays of those types)?
> 
> Do we need to care how they want to match an app to a device?
> 
> Is there value in us defining the specific parameters needed? There is a risk of needing to keep adding new things if we do define them.

Yes, we could also do it with a generic tag dictionary of key value pairs. Margo would move the correct matchmaking responsibility to their correct interaction. Then the details of the mchanisms of it would become out of scope for Margo. I understood the underlying feature request issue (https://github.com/margo/specification/issues/93) a bit differently, but I guess that would be fine too.

arne-broering:
> Pushing back on this SUP in its current shape. My concern is that it institutionalises vendor lock-in at the schema layer, which runs against Margo's stated purpose as a vendor-neutral interoperability layer.
> 
> The lock-in concentrates in three fields:
> 
> * runtimeRef keyed to a vendor-namespaced ID (com..) with no neutral equivalent
> * deviceSelector.vendorId: schema-level "must be this vendor"
> * deviceSelector.modelNumberIds[]: SKU-level gating
> 
> ...

I understand your concern but I guess it is inevitable that some app developers will create apps with very specific requirements on devices and actually focus vendor-specific runtimes. 
Still, the app developer can design the app for multiple ``targets`` (from different vendors) under the same ``deploymentProfile`` - this is supported by the proposal. So interoperability will be achieved this way.

I would be okay with moving this SUP's proposed mechanisms under some 'x-tension...' as described by @phil-abb in #56 . However, the result would be the same, only the governance of this might be moved out of Margo's core scope. Question is whether this would be a desriable outcome.

stormc:
Stepping back a bit, I'd like to generally comment on some statements in this thread (in random order):

> So this SUP essentially wants us to create a "requiredDuringSchedulingIgnoredDuringExecution" - basically: this app must run this device, or only on this device with this tag/label.

> An ApplicationDescription that targets these is, by construction, non-portable across Margo-compliant fleets

There must be a means (not saying by this exact SUP) to target a particular Margo Edge device having a specific set of capabilities with a particular application. Such an application simply needs these capabilities and can't run without those. Think of special hardware attached or configuration such as a bespoke configured real-time Linux Kernel. Hence, "Margo-compliant fleets" are not (necessarily) homogeneous and must not be. Not allowing suchlike is severely limiting the scopes and environments in which Margo could potentially be adopted in the field.

Now, whether you discriminate by a particularly crafted set of capabilities or upfront by a "device tag" is effectively the same, differing just in the mechanics.


> If multiple vendors adopt the pattern (and they will), the practical outcome is Margo certifying devices [...] 

Margo doesn't certify devices, i.e., you don't get a sticker to put on your box from Margo.


> [...] that can't run each other's apps, i.e. the pre-Margo status quo with a logo on it. 

They can and they will as long as the device capabilities are sufficient to run the application. This is the Margo interoperability promise. Granted, someone may be tempted to build a walled garden, which is something that you cannot prevent as Margo via technical means but the market will have to (re)solve.


> The legitimate matching need (capabilities, runtime versions) should be conformance-test-based, not vendor-identifier-based.

Effectively, a vendor ID + model + X represents a specific set of capabilities of a particular device, kind of like an alias. It's just syntactic sugar for a particular set of capabilities.


> Vendor-specific advertisement belongs in @phil-abb's vendor-extensions SUP, not the core schema.

I feel that the SUPs #55, #56, and #50 all tackle similar things on different ends and are somewhat interdependent. In general, all deal with how capabilities are to be described and where discrimination happens because of them.

That said, I do not have strong feelings about doing it one way or the other as long as the discussion revolves around "how" and not "if at all".

rejected/device-specific-runtime.md
```diff
@@ -0,0 +1,323 @@
...
+    requiredResources:
+      cpu:
+        cores: 2
+      memory: 2Gi
+    
+    # --- Proposed extension -
+    runtimeRequirements:
+      targets:
+        - runtimeRef: com.northstar-ida.industry.edge-os
+          version: ">=3.1.0 <4.0.0"
```
phil-abb, line 132: Do you have some ideas on how version comparisons would be done? The way organizations version their runtime is going to vary dramatically. Since we cannot control what people use for versions, it's not clear what approach should be used.

arne-broering: Idea for a solution would be to separate the version string and the comparison semantics. Something like:

```YAML
runtimeRef: com.northstar-ida.industry.edge-os
version: "3.1.1",
versionScheme: "semver" 
```

and options for `versionScheme` could be:

```
semver - semantic versioning (3.1.1)
semver-compatible - attempt semver parsing, fallback lexicographic
calver - calendar versioning  (2024.01)
firmware-style - e.g. V1.2B
numeric - simple numeric comparison
lexicographic - string comparison only
```

With this in place, a matching algorithm could first determine the scheme and then apply comparison.

phil-abb: Yeah, unfortunately, something more complex like this would be needed if version comparison is a requirement. Do we know if there are expectations for version comparison?

pauldbrooks: My provisional opinion (subject to change and Clark is RA's PM representative) is that this is a rabbit hole that we should stay out of.  I prefer to keep is simple and allow both device and app to provide a list (e.g. for a device designed for PaulsBrilliantOS v1.1; PetersGoodOS v3.2; MarysOKOS v7.6) and the Fleet Manager assume backwards compatibility but not forwards (e.g. an app asking for v3.6 can be assumed to run on a device v3.7 but cannot be assumed to run on a device v3.5).  And for revisions to be numeric (27) and not emotionally resonant (Golden Gate)

I have experience in different application space where revision matching gets really complex and critically important for users; it we want to go beyond the simple example then we need to brace ourselves for very complex examples that may take us places that we don't want to go....

arne-broering: > My provisional opinion (subject to change and Clark is RA's PM representative) is that this is a rabbit hole that we should stay out of. I prefer to keep is simple and allow both device and app to provide a list (e.g. for a device designed for PaulsBrilliantOS v1.1; PetersGoodOS v3.2; MarysOKOS v7.6) and the Fleet Manager assume backwards compatibility but not forwards (e.g. an app asking for v3.6 can be assumed to run on a device v3.7 but cannot be assumed to run on a device v3.5). And for revisions to be numeric (27) and not emotionally resonant (Golden Gate)
> 
> I have experience in different application space where revision matching gets really complex and critically important for users; it we want to go beyond the simple example then we need to brace ourselves for very complex examples that may take us places that we don't want to go....

Yes, keeping it simple in a first version is also fine with me.

phil-abb:
@margo/technical-wg 

This SUP to define support for device-specific runtimes is ready for final review. Please review the proposal and add any questions/comments to the PR.

phil-abb:
This SUP was rejected during voting, so I'm force-merging it in and moving it to the rejected folder at the same time: https://github.com/margo/specification-enhancements/issues/78

PR #58, opened 2026-05-15
WFM Identity Profile

matlec:
First draft of the WFM Identity Profile SUP for community feedback.

### DIFF
--- a/proposals/wfm-identity-profile.md
+++ b/proposals/wfm-identity-profile.md
@@ -1,0 +1,339 @@
+# Margo WFM Identity Profile and Management Interface Update
+
+- [Margo WFM Identity Profile and Management Interface Update](#margo-wfm-identity-profile-and-management-interface-update)
+  - [Owner](#owner)
+  - [Summary](#summary)
+  - [Reason for proposal](#reason-for-proposal)
+  - [Requirements alignment acknowledgement](#requirements-alignment-acknowledgement)
+  - [Technical proposal](#technical-proposal)
+    - [1. Scope and Structure](#1-scope-and-structure)
+    - [2. Terminology](#2-terminology)
+    - [3. Identity Model](#3-identity-model)
+      - [WFM identity](#wfm-identity)
+      - [WFM Client identity](#wfm-client-identity)
+      - [Identity representation](#identity-representation)
+      - [Recognition by the WFM](#recognition-by-the-wfm)
+      - [Recognition by the WFM Client](#recognition-by-the-wfm-client)
+    - [4. Provisioning (PR2)](#4-provisioning-pr2)
+      - [WFM provisioning](#wfm-provisioning)
+      - [WFM Client provisioning](#wfm-client-provisioning)
+    - [5. Lifecycle and Operator Playbook](#5-lifecycle-and-operator-playbook)
+    - [6. Authentication and Authorization at the WFM API](#6-authentication-and-authorization-at-the-wfm-api)
+      - [Authentication model](#authentication-model)
+      - [Canonical caller identity](#canonical-caller-identity)
+      - [Authorization](#authorization)
+      - [Scope: traffic-inspecting proxies](#scope-traffic-inspecting-proxies)
+    - [7. Application to the Margo Management Interface](#7-application-to-the-margo-management-interface)
+      - [Endpoints removed](#endpoints-removed)
+      - [Security scheme removed](#security-scheme-removed)
+      - [Endpoints retained, with `{clientId}` dropped](#endpoints-retained-with-clientid-dropped)
+    - [8. Security Considerations](#8-security-considerations)
+    - [9. Roadmap (Informative)](#9-roadmap-informative)
+  - [Alternatives considered](#alternatives-considered)
+    - [Treating the WFM Client as a generic workload identity](#treating-the-wfm-client-as-a-generic-workload-identity)
+    - [Keeping RFC 9421 HTTP Message Signatures](#keeping-rfc-9421-http-message-signatures)
+    - [Web PKI for the WFM TLS certificate](#web-pki-for-the-wfm-tls-certificate)
+  - [Rejection reason](#rejection-reason)
+
+## Owner
+
+[@matlec](https://github.com/matlec)
+
+## Summary
+
+This SUP defines the initial **Margo WFM Identity Profile** (building on the **[Margo Identity and Authorization Framework (MIAF)](./margo-identity-and-authorization-framework.md)**) — covering both the **WFM identity** and the **WFM Client identity** — and updates the **Margo Management Interface for Workload Fleet Management** to consume MIAF-aligned identities. A WFM gets a Trust-Domain-scoped identity that anchors its namespace; a WFM Client gets a Trust-Domain-scoped identity named under the WFM that issues it.
+
+The change replaces three PR1 mechanisms:
+
+- the WFM-assigned `clientId` carried in URL paths;
+- the per-WFM onboarding and certificate-distribution endpoints (`POST /api/v1/onboarding` and `GET /api/v1/onboarding/certificate`); and
+- the RFC 9421 HTTP Message Signatures security scheme used for client authentication.
+
+Authentication is mutual: a WFM Client presents its X.509-SVID and validates the WFM's SVID; caller identity at the WFM API is the authenticated WFM Client SPIFFE ID carried via mTLS.
+
+PR2 ships the identity model and the mTLS authentication surface. Automated enrollment — for both WFMs and WFM Clients — is deferred to PR3; see [§1](#1-scope-and-structure) for how PR2 implementations operate in the meantime, and the [Roadmap](#9-roadmap-informative) for what comes later.
+
+## Reason for proposal
+
+PR1 onboarding has each Client present its own X.509 certificate to the WFM, which pins that certificate (typically by fingerprint) and assigns a `clientId`. The WFM exposes its own TLS server-side trust anchor at `GET /api/v1/onboarding/certificate` for Clients to fetch and trust. Subsequent Client requests are authenticated by server-side TLS plus RFC 9421 HTTP Message Signatures keyed to the Client's certificate. That worked for PR1's scope, but it runs into structural problems under MIAF's Trust-Domain-scoped identity model.
+
+PR1's mechanisms are entirely per-WFM: the `clientId` namespace, the TLS trust anchor, and the onboarding flow all live at the WFM rather than at the Trust Domain. WFM Client identities are not portable across the Trust Domain; the WFM's per-endpoint trust anchor duplicates what MIAF's Trust Bundle does Trust-Domain-wide; and the onboarding flow exists only to bootstrap Clients into that per-WFM trust regime, with no equivalent need under MIAF.
+
+The WFM itself is also affected: PR1 gives it no Trust-Domain-scoped identity. This SUP names the WFM with a Trust-Domain-scoped SPIFFE ID and recognizes it as a first-class principal under MIAF — putting WFMs and WFM Clients on the same identity footing under one Trust Bundle.
+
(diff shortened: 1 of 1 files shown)


phil-abb:
@margo/technical-wg A new SUP has been started by @matlec. If anyone is interested in helping work on fleshing out the details of this SUP, please contact @matlec

phil-abb:
@margo/technical-wg This SUP to define the WFM identity profile as part of the MIAF is ready for final review. Please review the proposal and add any questions/comments to the PR.

phil-abb:
@matlec - Where do you see the device identity coming into play? This SUP proposes that the WFM client (and I would assume DFM client later) has its own identity that is seperate from the device identity. So, as I understand it, any communication between the WFM client and WFM service is only going to be using the WFM client identity and not the device identity.

matlec:
> @matlec - Where do you see the device identity coming into play? This SUP proposes that the WFM client (and I would assume DFM client later) has its own identity that is seperate from the device identity. So, as I understand it, any communication between the WFM client and WFM service is only going to be using the WFM client identity and not the device identity.

@phil-abb You are correct, having separate identities for devices, WFM clients, and later DFM clients aligns exactly with how I think about this. Per SUP, the WFM client would use its own (WFM client) identity to authenticate with the WFM, not the device identity. How the WFM client and device identity are related, is a decision I have deliberately deferred for a later SUP (see *Device identity profile* in [the MIAF SUP's roadmap](https://github.com/margo/specification-enhancements/blob/feat/miaf-sup/proposals/margo-identity-and-authorization-framework.md#work-deferred-beyond-pr2-)). I have also removed the device identity from the MIAF SUP itself, and deferred it to a later SUP.

chrisgclayton:
This is really coming along great and it is in line with what I wanted to see as well. I have a few questions to add in here.

1. I was wondering if we need to call out or put a link to the SPIFFE x509 SVID validation rules (i.e., exactly 1 SAN URI present)
2. The retrieval of the trust bundle leaves a question of "how to validate". Should we be stating that our expectation is that the trust bundle is rooted at a Root CA that is in the trusted CA list (if not it will always fail validation in strict validation rules will it not)?
3. Part of the validation of certificates is to check for revocation against OCSP or CRL. Have we considered if we need to add any language to indicate this should be done and that the OCSP or CRL must be accessible to the validator (easier on the WFM side, but will the WFM Client have access to the OCSP service if required)? If we have this it also improves the client revocation certificate path IMO (I see you have callout for client and mass revocation)
4. Should we think about putting in an identity segmentation for more granular instance identification. Consider an identity for mysfm that is something like (spiffe://mydomain/margo/wfm/mywfm). This identity we get from this and it makes a ton of sense because I know it is "mywfm". If I have multiple instances of mywfm running we could offer an identity like spiffe://mydomain/margo/wfm/mywfm/1 where "1" is the first instance. This could help when a node is compromised etc. and we could revoke just it etc.

matlec:
> This is really coming along great and it is in line with what I wanted to see as well. I have a few questions to add in here.
> 
> 1. I was wondering if we need to call out or put a link to the SPIFFE x509 SVID validation rules (i.e., exactly 1 SAN URI present)
> 2. The retrieval of the trust bundle leaves a question of "how to validate". Should we be stating that our expectation is that the trust bundle is rooted at a Root CA that is in the trusted CA list (if not it will always fail validation in strict validation rules will it not)?
> 3. Part of the validation of certificates is to check for revocation against OCSP or CRL. Have we considered if we need to add any language to indicate this should be done and that the OCSP or CRL must be accessible to the validator (easier on the WFM side, but will the WFM Client have access to the OCSP service if required)? If we have this it also improves the client revocation certificate path IMO (I see you have callout for client and mass revocation)
> 4. Should we think about putting in an identity segmentation for more granular instance identification. Consider an identity for mysfm that is something like (spiffe://mydomain/margo/wfm/mywfm). This identity we get from this and it makes a ton of sense because I know it is "mywfm". If I have multiple instances of mywfm running we could offer an identity like spiffe://mydomain/margo/wfm/mywfm/1 where "1" is the first instance. This could help when a node is compromised etc. and we could revoke just it etc.

Hey @chrisgclayton! Thanks for reviewing the SUP!

1. MIAF's [X.509-SVID profile](https://github.com/margo/specification-enhancements/blob/feat/miaf-sup/proposals/margo-identity-and-authorization-framework.md#x509-svid-profile) adopts the SPIFFE X.509-SVID spec by reference, which carries the "single SAN URI" requirement you mentioned, and the other structural constraints. I'd try to keep the Margo spec lean on that topic and not restate SPIFFE requirements to reduce any risk of both drifting. That said, a pointer in MIAF's [Certificate validation section](https://github.com/margo/specification-enhancements/blob/feat/miaf-sup/proposals/margo-identity-and-authorization-framework.md#certificate-validation) flagging that the referenced SPIFFE structural rules apply during validation seems worth adding to help implementers - I'll fold that in.
2. If you mean validating the TLS connection used to retrieve the Trust Bundle (rather than validating SVIDs against it once retrieved), that's the chicken-and-egg case MIAF calls out in [Initial trust bootstrap](https://github.com/margo/specification-enhancements/blob/feat/miaf-sup/proposals/margo-identity-and-authorization-framework.md#initial-trust-bootstrap). MIAF deliberately allows both (1) PKI-anchored validation (web / enterprise PKI), and (2) pinned trust. "Rooted at a Root CA in the trusted CA list" is covered by option 1, is this correct?
3. MIAF in PR2 ([deliberately](https://github.com/margo/specification-enhancements/blob/feat/miaf-sup/proposals/margo-identity-and-authorization-framework.md#work-deferred-beyond-pr2-)) defers standardization on an automated revocation protocol (yet), but instead offers [three options for revocations](https://github.com/margo/specification-enhancements/blob/feat/miaf-sup/proposals/margo-identity-and-authorization-framework.md#operator-revocation-playbook-pr2): (1) removal of identities from a relying-party managed allowlist, (2) trust bundle rotation (for mass revocations), (3) short SVID lifetimes. To your point, all of those options do not scale beyond PoC deployments :) And we absolutely revisit revocation in a follow-up SUP in PR3!
4. The good news is that the design already accommodates per-instance identification. `wfm-id` is treated as opaque and chosen by an operator. The "compromised node" case you describe is supported by assigning distinct `wfm-id`s per instance (e.g., `mywfm-1`, `mywfm-2`), so a single instance can be revoked by whatever revocation mechanism we decide upon after PR2 (see my answer to 3.). An operator who instead wants HA replicas behind a single logical WFM identity uses one shared `wfm-id` across replicas. Both shapes work today, but you're right that the SUP doesn't currently spell this out. I'll add a sentence to the terminology section making the operator-chosen granularity explicit. I'd be hesitant though to bake an instance segment into the path itself (e.g., `spiffe://.../wfm/mywfm/1`). That would mandate a specific deployment shape (one logical WFM with N instances) and complicate the `wfm-id` namespace check that defends against the ["wrong-WFM Client acceptance" threat](https://github.com/margo/specification-enhancements/blob/feat/miaf-wfm-identity-sup/proposals/wfm-identity-profile.md#8-security-considerations). Does this cover what you had in mind?

chrisgclayton:
> > This is really coming along great and it is in line with what I wanted to see as well. I have a few questions to add in here.
> > 
> > 1. I was wondering if we need to call out or put a link to the SPIFFE x509 SVID validation rules (i.e., exactly 1 SAN URI present)
> > 2. The retrieval of the trust bundle leaves a question of "how to validate". Should we be stating that our expectation is that the trust bundle is rooted at a Root CA that is in the trusted CA list (if not it will always fail validation in strict validation rules will it not)?
> > 3. Part of the validation of certificates is to check for revocation against OCSP or CRL. Have we considered if we need to add any language to indicate this should be done and that the OCSP or CRL must be accessible to the validator (easier on the WFM side, but will the WFM Client have access to the OCSP service if required)? If we have this it also improves the client revocation certificate path IMO (I see you have callout for client and mass revocation)
> > 4. Should we think about putting in an identity segmentation for more granular instance identification. Consider an identity for mysfm that is something like (spiffe://mydomain/margo/wfm/mywfm). This identity we get from this and it makes a ton of sense because I know it is "mywfm". If I have multiple instances of mywfm running we could offer an identity like spiffe://mydomain/margo/wfm/mywfm/1 where "1" is the first instance. This could help when a node is compromised etc. and we could revoke just it etc.
> 
> Hey @chrisgclayton! Thanks for reviewing the SUP!
> 
> 1. MIAF's [X.509-SVID profile](https://github.com/margo/specification-enhancements/blob/feat/miaf-sup/proposals/margo-identity-and-authorization-framework.md#x509-svid-profile) adopts the SPIFFE X.509-SVID spec by reference, which carries the "single SAN URI" requirement you mentioned, and the other structural constraints. I'd try to keep the Margo spec lean on that topic and not restate SPIFFE requirements to reduce any risk of both drifting. That said, a pointer in MIAF's [Certificate validation section](https://github.com/margo/specification-enhancements/blob/feat/miaf-sup/proposals/margo-identity-and-authorization-framework.md#certificate-validation) flagging that the referenced SPIFFE structural rules apply during validation seems worth adding to help implementers - I'll fold that in.
> 2. If you mean validating the TLS connection used to retrieve the Trust Bundle (rather than validating SVIDs against it once retrieved), that's the chicken-and-egg case MIAF calls out in [Initial trust bootstrap](https://github.com/margo/specification-enhancements/blob/feat/miaf-sup/proposals/margo-identity-and-authorization-framework.md#initial-trust-bootstrap). MIAF deliberately allows both (1) PKI-anchored validation (web / enterprise PKI), and (2) pinned trust. "Rooted at a Root CA in the trusted CA list" is covered by option 1, is this correct?
> 3. MIAF in PR2 ([deliberately](https://github.com/margo/specification-enhancements/blob/feat/miaf-sup/proposals/margo-identity-and-authorization-framework.md#work-deferred-beyond-pr2-)) defers standardization on an automated revocation protocol (yet), but instead offers [three options for revocations](https://github.com/margo/specification-enhancements/blob/feat/miaf-sup/proposals/margo-identity-and-authorization-framework.md#operator-revocation-playbook-pr2): (1) removal of identities from a relying-party managed allowlist, (2) trust bundle rotation (for mass revocations), (3) short SVID lifetimes. To your point, all of those options do not scale beyond PoC deployments :) And we absolutely revisit revocation in a follow-up SUP in PR3!
> 4. The good news is that the design already accommodates per-instance identification. `wfm-id` is treated as opaque and chosen by an operator. The "compromised node" case you describe is supported by assigning distinct `wfm-id`s per instance (e.g., `mywfm-1`, `mywfm-2`), so a single instance can be revoked by whatever revocation mechanism we decide upon after PR2 (see my answer to 3.). An operator who instead wants HA replicas behind a single logical WFM identity uses one shared `wfm-id` across replicas. Both shapes work today, but you're right that the SUP doesn't currently spell this out. I'll add a sentence to the terminology section making the operator-chosen granularity explicit. I'd be hesitant though to bake an instance segment into the path itself (e.g., `spiffe://.../wfm/mywfm/1`). That would mandate a specific deployment shape (one logical WFM with N instances) and complicate the `wfm-id` namespace check that defends against the ["wrong-WFM Client acceptance" threat](https://github.com/margo/specification-enhancements/blob/feat/miaf-wfm-identity-sup/proposals/wfm-identity-profile.md#8-security-considerations). Does this cover what you had in mind?

1. Perfect, thanks
2. I meant the root certificate downloaded in the trust bundle that is required to complete the chain. Unfortunately I just reread the SPIFFE trust bundle documentation and they do not speak to it. I guess we could do the same and either indicate the device needs to validate the chain and possibly call out the point?
3. Ok, I will flag a revisit ;)
4. Yes this should be fine but not calling out leaves us flexible I agree. The disadvantage is that we cannot assume WFM Client and WFM Server will know how to make sense of these and differentiate. I was just wondering if we should say something more opinionated on it such as spiffe://.../wfm/{id}/{optional} id will always be one and exactly one entry between the slash after wfm. There is an optional pathing for more granular identification (I am ok with your approach if preferred)

phil-abb:
Merging in this PR since the SUP has been approved: https://github.com/margo/specification-enhancements/issues/71#issuecomment-4768116513

PR #56, opened 2026-05-01
New SUP for specification extensions

phil-abb:
A new SUP for the feature to support vendor extensions: https://github.com/margo/specification/issues/144

PR #59, opened 2026-05-19
Alternative SUP for device specific runtimes

phil-abb:


phil-abb:
@margo/technical-wg An alternative SUP for device-specific runtime support has been started by @phil-abb. If anyone is interested in helping work on fleshing out the details of this SUP, please contact @phil-abb 

This is an alternative approach to the [device specific runtimes SUP](https://github.com/margo/specification-enhancements/pull/55) started by @arne-broering

phil-abb:
One area this SUP doesn't address yet is the conflict/confusion adding the deviceAffinity expressions has with also specifying the [requiredResources](https://docs.margo.org/specification/applications/application-description#requiredresources-attributes) in the Application Description file.

This needs to be clarified in the SUP.

I think the peripherals and interfaces could easily be handled because they fit in the property selector. The CPU, Memory, and Storage are a bit different because the matching expectations are a little different for these.

chrisgclayton:
As I read through this I was questioning if the device affinity was really device affinity vs. placement constraints? Normally affinity IME is around having things "be drawn together" but it seems more like placement constraints in usage? Also do we need the opposite to be true? For example if we are thinking of these as placement constraints etc. saying "target where label MyLabel == 'MyValue'" do we need to also have the ability to say, do not place it on any node that has "MyOtherLabel == 'MyOtherValue'"?

arne-broering:
@phil-abb: Thanks for thinking this further. You nicely elaborated the semantics and gave very detailed examples. 

I guess the mechanics around allowing device-specific runtimes are similar to what is proposed in my earlier SUP #55 with the 2 additional elements on DeviceCapabilitiesManifest and ApplicationDescription. Though, this approach relates to much more than just device-specific runtimes. The ``deviceAffinity`` mechanism on ApplicationDescription is a very powerful tool for app developers. And it is highly flexible compared to the `runtimeRequirements` key that is in #55. I think this is a good idea. 

You could actually extend it towards filtering over your `x-tension` approach in #56 as well. Then, I wonder why we need the ``label`` (as proposed in this SUP) in DeviceCapabilitiesManifest (containing vendor defined keys and values) if we anyways have the `x-tension` (#56) at the same location in the DeviceCapabilitiesManifest. So the given example [1] can be expressed as [2] by means of #56 mechanism:

Example [1]:
```json
{
    "properties": {...},
    "labels": {
      "example.com/hyper-v.host": true 
    }
}
```

Example [2] using #56 `x-tension` mechanism:
```json
{
    "properties": {...},
    "x-hyper-v-extension": {
      "example.com/hyper-v.host": true 
    }
}
```

No need for `label` right?

Now, I see that what I defined under #55 to extend the ``DeviceCapabilitesManifest`` with a `runtimeEnvironments` element could be put inside an `x-tension` element (similar to what is shown in [2]) and with ``deviceAffinity`` we could still match on that.

pauldbrooks:
@chrisgclayton I see a difference between the device side and the workload side. A device has an attribute that some apps may prefer to use. it simply unemotionally documents what it has; the device does not know which apps may want or need to use it. The app/workload needs to be able to inform the Fleet Manager if it requires a specific attribute to be present.  But you raise another interesting question; an app may prefer to be used in conjunction with a specific capability but is able to work around it, with consequences that the user may not care about. None of this changes a fundamental requirement on the app; it must validate that the device is capable when it boots up - it cannot rely on the fleet manager whose primary role in this scenario is to deliver ease of use.

pauldbrooks:
@arne-broering in your example, is the list of properties freeform, allowing me to introduce a property called "family-brooks.org/immos-tongue" (you're not meant to understand it) or would it require all properties to be defined in a Margo lexicon?

arne-broering:
> @arne-broering in your example, is the list of properties freeform, allowing me to introduce a property called "family-brooks.org/immos-tongue" (you're not meant to understand it) or would it require all properties to be defined in a Margo lexicon?

@pauldbrooks yes, what you would write under a key `x-...` is freeform. @phil-abb has proposed this in #56 - so currently under discussion.

phil-abb:
> As I read through this I was questioning if the device affinity was really device affinity vs. placement constraints? Normally affinity IME is around having things "be drawn together" but it seems more like placement constraints in usage? 

I'm not too passionate about naming, so if something other than "deviceAffinity" works better, I'm not opposed to it. Considering this is stolen directly from what Kubernetes is doing with node/pod affinity, it seemed like the term fit to me because it's pretty much doing the same thing; expect the WFM is acting as the "scheduler".  I'd be interested in hear other opinions on the use of "deviceAffinity" or some other term. 

>Also do we need the opposite to be true? For example if we are thinking of these as placement constraints etc. saying "target where label MyLabel == 'MyValue'" do we need to also have the ability to say, do not place it on any node that has "MyOtherLabel == 'MyOtherValue'"?

Potentially. I haven't heard any use cases yet where there is a desire not to deploy an app to a device if it has certain qualities, only if it does have them. If anyone has any use cases where they would need to say, " Don't deploy my app to a device that has XYZ", I'd be interested in hearing about it. If it makes sense to have "affinity" and "anti-affinity" (or whatever we end up calling it), we can add it, but I don't want to add it just to have it.

phil-abb:
>Example [2] using https://github.com/margo/specification-enhancements/pull/56 x-tension mechanism:

>
>{
>    "apiVersion": "device.margo.org/v1alpha1",
>    "kind": "DeviceCapabilitiesManifest",
>   "properties": {...},
>    "x-hyper-v-extension": {
>     "example.com/hyper-v.host": true 
>    }
> }
> No need for label right?

I'm not a fan of this because it infers meaning to what is under "x-\<company-name\>-extensions." The intention for the specification extensions is to allow suppliers to add anything they want (as long as it's valid JSON/YAML) to this document, and Margo doesn't ascribe any meaning at all to it. It only has meaning to the suppliers using it.

"labels" have a purpose and meaning in the Margo specification. It's used specifically for allowing suppliers to provide arbitrary information that is used for identification purposes as a counter to the "properties", which is Margo's prescribed identifying information.

phil-abb:
> One area this SUP doesn't address yet is the conflict/confusion adding the deviceAffinity expressions has with also specifying the [requiredResources](https://docs.margo.org/specification/applications/application-description#requiredresources-attributes) in the Application Description file.
> 
> This needs to be clarified in the SUP.
> 
> I think the peripherals and interfaces could easily be handled because they fit in the property selector. The CPU, Memory, and Storage are a bit different because the matching expectations are a little different for these.

The more I think about it, the more I'm not a fan of how we have all of these grouped under "resources" in the device capabilities. CPU, Memory, and Storage are not the same things as Peripherals and Interfaces. CPU, Memory, and Storage are allocatable/consumable resources that deal with compute. Peripherals and Interfaces are more general capabilities. They either exist on the device or they don't. I could see splitting this out so we could have a requiredCompute in the application definition that's just targeting CPU, Memory, and Storage, and then allow for Peripheral and Interfaces to be matched using the propertySelector.

This is also an issue I'm starting to have with the [SUP to remove the roles](https://github.com/margo/specification-enhancements/blob/sup_roles_to_capabilities/proposals/sup_device_roles_to_capabilities.md). We are creating a "capabilities" category, but the document itself is device capabilities, so it seems redundant. I also think the Peripheral and Interfaces should be at the same level/places as what the sup is putting under "capabilities." I think all of these should be at the "properties" root.

@ajcraig what do you think?

Or maybe the file shouldn't even be called deviceCapabilities because things like vendor, serial number, and model are not capabilities. Maybe it should be deviceDescription instead - but I'll defer this line of thinking until another time.

phil-abb:
> > As I read through this I was questioning if the device affinity was really device affinity vs. placement constraints? Normally affinity IME is around having things "be drawn together" but it seems more like placement constraints in usage?
> 
> I'm not too passionate about naming, so if something other than "deviceAffinity" works better, I'm not opposed to it. Considering this is stolen directly from what Kubernetes is doing with node/pod affinity, it seemed like the term fit to me because it's pretty much doing the same thing; expect the WFM is acting as the "scheduler". I'd be interested in hear other opinions on the use of "deviceAffinity" or some other term.
> 

A couple of other options instead of "deviceAffinity":
- deviceConstraints
- deviceRequirements
- capabilityRequirements
- deploymentConstraints
- deploymentRequirements
- targetConstraints
- placementConstraints

"deploymentConstraints" might fit best with it being part of the deployment profile. Or, since it's already under "deploymentProfiles," it could just be "constraints."

@chrisgclayton

phil-abb:
> One area this SUP doesn't address yet is the conflict/confusion adding the deviceAffinity expressions has with also specifying the [requiredResources](https://docs.margo.org/specification/applications/application-description#requiredresources-attributes) in the Application Description file.
> 
> This needs to be clarified in the SUP.
> 
> I think the peripherals and interfaces could easily be handled because they fit in the property selector. The CPU, Memory, and Storage are a bit different because the matching expectations are a little different for these.

The problem with the current [requiredResources](https://docs.margo.org/specification/applications/application-description#requiredresources-attributes) is that it only matches with CPU, Memory, Storage, Peripherals, and Interfaces. It doesn't match on vendor, serial number, or model number. This is why the "propertySelector" was added to this SUP. The propertySelector could be used to match memory, storage, and interfaces since they are simple objects. The problem is the complex resources CPU and Peripherals, since you would want to match on multiple things (e.g., type=GPU & manufacturer=NVIDIA).  This could be handled with some enhancements to the matchExpression:

```yaml
deviceAffinity:
  - propertySelector:
      matchExpressions:
        - key: /resources/peripherals
          operator: All
          itemSelector:
            matchExpressions:
              - key: /type
                operator: In
                values: ["GPU"]
              - key: /manufacturer
                operator: In
                values: ["NVIDIA"]
```

phil-abb:
@jjaswanson4 This is an alternative device-specific runtime SUP. I think it aligns better with what you commented on in the other SUP: https://github.com/margo/specification-enhancements/pull/55#issuecomment-4338291412

I'd be interested in your thoughts on this.

chrisgclayton:
> @chrisgclayton I see a difference between the device side and the workload side. A device has an attribute that some apps may prefer to use. it simply unemotionally documents what it has; the device does not know which apps may want or need to use it. The app/workload needs to be able to inform the Fleet Manager if it requires a specific attribute to be present. But you raise another interesting question; an app may prefer to be used in conjunction with a specific capability but is able to work around it, with consequences that the user may not care about. None of this changes a fundamental requirement on the app; it must validate that the device is capable when it boots up - it cannot rely on the fleet manager whose primary role in this scenario is to deliver ease of use.

Agreed

chrisgclayton:
> > As I read through this I was questioning if the device affinity was really device affinity vs. placement constraints? Normally affinity IME is around having things "be drawn together" but it seems more like placement constraints in usage?
> 
> I'm not too passionate about naming, so if something other than "deviceAffinity" works better, I'm not opposed to it. Considering this is stolen directly from what Kubernetes is doing with node/pod affinity, it seemed like the term fit to me because it's pretty much doing the same thing; expect the WFM is acting as the "scheduler". I'd be interested in hear other opinions on the use of "deviceAffinity" or some other term.
> 
> > Also do we need the opposite to be true? For example if we are thinking of these as placement constraints etc. saying "target where label MyLabel == 'MyValue'" do we need to also have the ability to say, do not place it on any node that has "MyOtherLabel == 'MyOtherValue'"?
> 
> Potentially. I haven't heard any use cases yet where there is a desire not to deploy an app to a device if it has certain qualities, only if it does have them. If anyone has any use cases where they would need to say, " Don't deploy my app to a device that has XYZ", I'd be interested in hearing about it. If it makes sense to have "affinity" and "anti-affinity" (or whatever we end up calling it), we can add it, but I don't want to add it just to have it.

Just to explain why my concern with device affinity as a term consider the following user scenarios. 

1. Does the WFM say "deploy vision inspector to qadevice1" and then these are used to say "vision inspector requires a Onvif pivot capable capable camera but qadevice1 only has a stationary camera that does not speak Onvif so you cannot deploy there.
2. Does the WFM say "deploy vision inspector to a device that supports Onvif pivot capable cameras"

In scenario #1 this does not really align to the concept of Kubernetes as I have already decided where to deploy it and this is a "constraint". 

In scenario #2 this does align to your point with Kubernetes and I am saying here is a set of devices I want to push this to, make sure it is deployed to the place that can support it.

I am ok if we want to keep it the way it is, just wanted to raise it as it feels odd to me

pauldbrooks:
Digging into @chrisgclayton example, because I think it brings out the ultimate _responsibility_ of WFM and workload.  

- Let's assume that "qadevice1" always ships with an "Onvif pivot capable capable camera", and that "vision inspector" is pretested with that configuration.  It is reasonable for "vision inspector" to have affinity with "qadevice1".
- Now let's assume that "Onvif pivot capable capable camera" has broken and been replaced with a generic stationary camera.
- There is no reason for the WFM to be aware of that.  The workload can still be deployed.
- It should be the workload's responsibility to validate that "Onvif pivot capable capable camera" is installed, working and available.  And if not, then the workload should be sending telemetry to explain why it's not working.
- Now let's assume that the natural human being driving the WFM has "randomedgedevice" which is fitted with "Onvif pivot capable capable camera" and wants to use it
- It is reasonable for the workload to not permit that (and not run if forced onto the device)
- It is reasonable for the workload to also have affinity with edge devices that claim the "Onvif pivot capable capable camera" capability. And it is reasonable for that workload to want the human being to see a warning "Untested hardware platform, please test functionality before using in a production environment". And it's reasonable for the workload to provide telemetry saying "Running on untested edge device - performance cannot be guaranteed"

My takeaways from this thought process

- Affinity must be defined by a list of complex AND/OR/XOR expressions
- Affinity may benefit from being ranked
- Affinity must be validated by the workload when deployed on an edge device
- Affinity is an ease of use tool at WFM and we need to carefully consider how far to take conformity

phil-abb:
> Just to explain why my concern with device affinity as a term consider the following user scenarios.
> 
> 1. Does the WFM say "deploy vision inspector to qadevice1" and then these are used to say "vision inspector requires a Onvif pivot capable capable camera but qadevice1 only has a stationary camera that does not speak Onvif so you cannot deploy there.
> 2. Does the WFM say "deploy vision inspector to a device that supports Onvif pivot capable cameras"
> 
> In scenario #1 this does not really align to the concept of Kubernetes as I have already decided where to deploy it and this is a "constraint".
> 
> In scenario #2 this does align to your point with Kubernetes and I am saying here is a set of devices I want to push this to, make sure it is deployed to the place that can support it.
> 
> I am ok if we want to keep it the way it is, just wanted to raise it as it feels odd to me

Ultimately, it will be up to the device to determine if the workload attempting to be deployed there can run there. However, we want to enable the WFM to enable functionality to allow customers to make smart decisions about where applications are deployed to limit deployment failures. So in this case, it's kind of both, but we're trying to enable scenario #1.

I'll go ahead and work on changing the property names just to make sure there are no automatic assumptions made about functionality based on the naming.

phil-abb:
>Just to explain why my concern with device affinity as a term, consider the following user scenarios.

@chrisgclayton I pushed up some updates that change the property names.

phil-abb:
> > One area this SUP doesn't address yet is the conflict/confusion adding the deviceAffinity expressions has with also specifying the [requiredResources](https://docs.margo.org/specification/applications/application-description#requiredresources-attributes) in the Application Description file.
> > This needs to be clarified in the SUP.
> > I think the peripherals and interfaces could easily be handled because they fit in the property selector. The CPU, Memory, and Storage are a bit different because the matching expectations are a little different for these.
> 
> The problem with the current [requiredResources](https://docs.margo.org/specification/applications/application-description#requiredresources-attributes) is that it only matches with CPU, Memory, Storage, Peripherals, and Interfaces. It doesn't match on vendor, serial number, or model number. This is why the "propertySelector" was added to this SUP. The propertySelector could be used to match memory, storage, and interfaces since they are simple objects. The problem is the complex resources CPU and Peripherals, since you would want to match on multiple things (e.g., type=GPU & manufacturer=NVIDIA). This could be handled with some enhancements to the matchExpression:
> 
> ```yaml
> deviceAffinity:
>   - propertySelector:
>       matchExpressions:
>         - key: /resources/peripherals
>           operator: All
>           itemSelector:
>             matchExpressions:
>               - key: /type
>                 operator: In
>                 values: ["GPU"]
>               - key: /manufacturer
>                 operator: In
>                 values: ["NVIDIA"]
> ```

Updates were made to handle capacity (memory, storage, RAM) and object array (peripherals) matching

phil-abb:
@margo/technical-wg 

This alternative SUP to define support for device-specific runtimes is ready for final review. Please review the proposal and add any questions/comments to the PR. This is an alternative approach to [this SUP](https://github.com/margo/specification-enhancements/pull/55).

proposals/sup_device_specific_runtime_affinity_matching.md
```diff
@@ -0,0 +1,893 @@
...
+* `values` MUST be parsable as numbers
+* `values` MUST only contain a single number
+
+For `ContainsAll` and `ContainsAny`:
+
+* `itemSelector` MUST be present
+* `itemSelector.matchExpressions` MUST contain one or more expressions
+* `values` MUST be omitted
+* Keys within `itemSelector.matchExpressions` are JSON Pointers relative to each array element (not absolute)
+* Expressions within `itemSelector.matchExpressions` are combined using AND logic
```
matlec: This line contradicts the definition of `ContainsAny` below. There we say that `ContainsAny` combines the `itemSelector` expressions using **OR**, but this line applies **AND** to both operations (which makes `ContainsAny` identical to `ContainsAll`).

```suggestion
* Expressions within `itemSelector.matchExpressions` are combined using `AND` logic for `ContainsAll` and `OR` logic for `ContainsAny`
```

phil-abb:
This SUP has been approved, so force-merging the SUP: https://github.com/margo/specification-enhancements/issues/78

PR #65, opened 2026-05-29
The Helm Way. Better Compose and Quadlet logistics

javatask:
The Margo specification already provides a mature, OCI-native publishing workflow for Helm-based components. This document names that workflow "The Helm Way" and distills it into five reusable elements so that future component types (Compose, Quadlet, or any type added later) can reference a single, well-defined pattern rather than repeating the same requirements in every proposal.

### DIFF
--- a/proposals/the-helm-way/PATCH_desired-state-quadlet.diff
+++ b/proposals/the-helm-way/PATCH_desired-state-quadlet.diff
@@ -1,0 +1,108 @@
+# PATCH: desired-state.linkml.yaml -- Add QuadletDeploymentProfile and QuadletComponent
+#
+# Schema file:
+#   src/specification/margo-management-interface/desired-state.linkml.yaml
+#
+# Purpose:
+#   Close R3 gap identified by margo-tech-reviewer in REVIEW_tech-reviewer.md.
+#   Without these additions, WFMs cannot generate schema-valid
+#   ApplicationDeployment desired-state documents for Quadlet workloads.
+#   The type slot accepts the string "quadlet" (no pattern constraint exists
+#   in the desired-state schema) but there is no QuadletDeploymentProfile or
+#   QuadletComponent class to validate against, leaving Quadlet desired-state
+#   documents effectively untyped.
+#
+# How to apply:
+#   This unified diff is NOT committed directly to the submodule. To apply:
+#
+#       cd documentation/spec_analysis/specification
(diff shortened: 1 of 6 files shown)


phil-abb:
@javatask, couple of comments:

* I noticed this PR has several SUPs and not just the "Helm Way" SUP. I would suggest only including the files for a single SUP with each PR, so it's clear where to focus attention/conversation. Also, each SUP will have its own lifecycle, so by combining things together, it makes it harder to handle them individually (e.g., moving from pre-draft to draft, figuring out when things are ready for voting, moving to completed, etc.)
* I noticed you didn't use the [SUP Template](https://github.com/javatask/specification-enhancements/blob/the-helm-way/sup-template.md). It's not that big of a deal because your template covers most sections, but I would suggest you at least add the `Requirements alignment acknowledgement` section to your documents because each SUP needs to have a companion feature that has been agreed is within scope.
* I'm not sure what your intentions are, but there are [two stages](https://github.com/margo/specification-enhancements#sup-status) before voting on SUPs:
  * Phase 3: SUP Technical Design (P1) - This means there is an intention to continue to do some work on the SUP, and the SUP owner doesn't feel it's complete yet. Or maybe the SUP owner wants to give some time for the community to provide some initial feedback before indicating they think it's ready to vote on.
  * Phase 3: Final Review (P2) - This means the SUP owner feels the content is complete and ready to be voted on.
  * If you feel the SUP is still in P1, then please make this PR a draft to indicate this is the case
  * If you feel the SUP is ready for P2, then keep it as is. This means the two-week timer will start today for when the SUP becomes eligible for voting. Voting will be scheduled once it looks like all community comments are addressed in some way (may take longer than 2 weeks).
  * Please let me know which of these is correct so I can send the proper message.
  
  
  You can find the detailed information about [the SUP process here](https://github.com/margo/Margo-collaboration/blob/main/specification_update_processes/process-overview.md)

phil-abb:
@javatask / @ajcraig 

Even though the WG-PROPOSAL-00: The Helm Way SUP isn't something that impacts the specification/documentation directly, it's still setting a precedent for how we work moving forward, so I think it makes sense to still vote on this. 

It's a little bit different than the previous SUPs since it's not focused on something that changes the spec, just the way we work, so maybe there is some room to handle the voting on this a bit differently? Or, include it as an implication of voting on the Quadlet or Better Compose SUP? 

I personally like this idea and would rather see it voted on on its own instead of tying it to another SUP that might get rejected for some other reason.

PR #66, opened 2026-05-29
[DRAFT] Resource discovery and conflict resolution proposal for app workloads

singhmj-1:
Addresses:  https://github.com/margo/specification/issues/75


phil-abb:
@singhmj-1 

I'm not sure what your intentions are, but there are [two stages](https://github.com/margo/specification-enhancements#sup-status) before voting on SUPs:

* Phase 3: SUP Technical Design (P1) - This means there is an intention to continue to do some work on the SUP, and the SUP owner doesn't feel it's complete yet. Or maybe the SUP owner wants to give some time for the community to provide some initial feedback before indicating they think it's ready to vote on.
* Phase 3: Final Review (P2) - This means the SUP owner feels the content is complete and ready to be voted on.

If you feel the SUP is still in P1, then please make this PR a draft to indicate this is the case

If you feel the SUP is ready for P2, then keep it as is. This means the two-week timer will start today for when the SUP becomes eligible for voting. Voting will be scheduled once it looks like all community comments are addressed in some way (may take longer than 2 weeks).

Please let me know which of these is correct so I can send the proper message.

singhmj-1:
@phil-abb this is P1, and I've marked it as draft. Thanks.

phil-abb:
@margo/technical-wg A new SUP resource allocation and conflict resolution has been started by @singhmj-1 . If anyone is interested in helping work on fleshing out the details of this SUP, please contact @singhmj-1 .

phil-abb:
@singhmj-1 / @arne-broering 

I can see some potential collaboration between the ideas presented in this SUP and the [Device-specific parameters SUP](https://github.com/margo/specification-enhancements/pull/54). It seems like the ideas for how the device-specific parameters SUP is proposing the Application Definition indicate it wants values provided from somewhere other than a user or default, fits in with what this SUP wants to accomplish.

phil-abb:
I'm not quite following how the provider types are used. The SUP describes the schema, but it doesn't indicate what is done with the schema. There seems to be some steps missing.

 I'm assuming the device creates instances of one or more of these provider-type documents, based on the schema that is specific to the device, and somehow sends this to the WFM so the WFM can use it while creating the application deployment YAML files?

singhmj-1:
Initially I had the same thought, but `providers` do not need to be published by the `device`. The WFM will maintain the providers list, and can decouple the implementation details (whether sourced via device or an internal WFM mechanism). The manifest only requires the resolved parameter value, it doesn't matter whether it discover these values at WFM or device.

Example Flow 1 (_capability.margo.org/network/port_): WFM implements the provider internally. WFM provisions the Network Port on the device using proprietary APIs. WFM logs the resource in its local database. WFM injects the port name into the application deployment manifest. The entire manifest gets rendered on WFM, and then sent over to device.

Example Flow 2(_capability.margo.org/network/port_): WFM injects the `infrastructureBindings` definitions in the desired state. Then the device will be taking care of how the resolution is done for these bindings. The entire manifest gets rendered on the device.

phil-abb:
> Initially I had the same thought, but `providers` do not need to be published by the `device`. The WFM will maintain the providers list, and can decouple the implementation details (whether sourced via device or an internal WFM mechanism). The manifest only requires the resolved parameter value, it doesn't matter whether it discover these values at WFM or device.

Sorry, I'm still not following. 

Take for example:

* capability.margo.org/network/port - how would the WFM know what ports the device is OK with using or what strategy it wants people to use?
* capability.margo.org/camera - How would the WFM know what interfaces or resolution the camera has?

Maybe I just need to see an example of the Application Deployment YAML for the case where the WFM isn't providing the values, because the example shows the resolved values coming from the WFM.

singhmj-1:
@phil-abb You are right. After pondering on this, I agree that this responsibility should be handed over to the device. Will need to update Capabilities spec so that device can broadcast what `providers` it supports, for this SUP to work.

Meanwhile, I would like to give you context on what I was originally picturing for the WFM-side resolution: I was imagining a scenario where the WFM sees `capability.margo.org/network/port` in an app, hits a device's proprietary API (e.g., /available-ports), receives a list of available ports back, and then the WFM picks one to seed into the manifest. My goal with that logic was to keep the implementation details fully decoupled and leave it up to WFM and device vendors to decide how to handle it. However, your point made me realize that I was introducing non-interoperability.

**P.S. :** I still believe resource resolution remains valid at both the WFM and device levels, **depending entirely on the implementation scope**. A clear example of WFM-level resolution is seeding sensitive enterprise configurations or credentials into the manifest. For security and isolation reasons, the edge device has no direct visibility into upstream cloud control planes or secret managers (like AWS Secrets Manager or Vault). The WFM is the only entity with the connectivity and IAM permissions to fetch these objects, meaning it must resolve the values at the orchestration layer before passing the finalized parameters down to the device:

```yaml
apiVersion: margo.org/v1-alpha1
kind: DeploymentManifest
metadata:
  id: deployment-test-app-device-xyz-002
  applicationId: test-app-123456780

###############################################################################
infrastructureBindings:
  # Resolved entirely at the WFM layer because the edge device lacks 
  # direct cloud access/credentials to talk to AWS.
  - id: databaseCredentials
    provider: capability.margo.org/secret/aws-vault
    input:
      secretName: my-database-creds

# Bindings resolve to ApplicationDescription parameter names
parameters:
  - name: databaseUsername
    fromBinding: databaseCredentials.output.username

  - name: databasePassword
    fromBinding: databaseCredentials.output.password

  - name: databaseHostname
    fromBinding: databaseCredentials.output.hostname
```

proposals/resource-allocation-to-workloads.md
```diff
@@ -0,0 +1,446 @@
...
+schema:
+  input:
+    preferredPort:
+      type: integer
+      minimum: 1
+      maximum: 65535
+      description: Preferred port number; treated as advisory unless strategy is strict.
+    protocol:
+      type: enum
+      values: [TCP, UDP]
```
jachstet-sea: I propose to add SCTP here

singhmj-1: Thanks for pointing that out, @jachstet-sea! I just threw that example together to help illustrate the baseline concept. The main goal here is to introduce the providers framework so it can be extended later. Appreciate the feedback! :)

singhmj-1: Always open to better examples if you have any ideas! : )

proposals/resource-allocation-to-workloads.md
```diff
@@ -0,0 +1,446 @@
...
+Device Agent                       (applies resolved values to deployment components)
+```
+
+---
+
+### Predefined Provider Types and their Formal Schemas
+
+#### `capability.margo.org/network/port`
+
+```yaml
```
nilanjan-samajdar: If the intent is to declare port-ranges in the device available to Applications, the format string should allow a list of ports as well as port ranges. For example - 
`
[
  { start: 1024,
    end: 49151 }
]
` 

or

`
[
  2345, 
   4095, 
   { start: 8080, 
     end: 49150}
]
`

nilanjan-samajdar (commented):
Have some suggestions to allow the device to communicate the available ports for applications as a combination of specific port numbers as well as port ranges.

singhmj-1:
I am currently working on formally introducing the providers concept directly into this SUP. I will be rewriting portions and introduce some new text to bring more clarity and structure to the overall proposal. Right now, the draft doesn't properly define what providers are, their operational scope, how they are introduced into the system, and the specific problems they solve for us.

I will drop a comment here once I have all the updates in place for another look. Thanks, everyone!

cc: @phil-abb @jachstet-sea @nilanjan-samajdar

phil-abb:
@singhmj-1 

> I was imagining a scenario where the WFM sees capability.margo.org/network/port in an app, hits a device's proprietary API (e.g., /available-ports), receives a list of available ports back, and then the WFM picks one to seed into the manifest. 

Keep in mind that communication can only originate from the device to the WFM and not from the WFM to the device.

> I will be rewriting portions and introduce some new text to bring more clarity and structure to the overall proposal.

Sound good.

proposals/resource-allocation-to-workloads.md
phil-abb: I think this SUP would benefit from having some documented use cases that highlight the problem this SUP is attempting to solve. 

For example, with the `capability.margo.org/network/ingress` provider type, it's not clear to me how all this information is supposed to be used. 

One of the cases I have:

I have a Kubernetes application that defines a default ingress route root of `hello-world.\<hostname\>` (e.g., https://hello-world.device1.example.com/say-hello). Since I'm providing the ingress route manifest in my deployment, the protocol and tlsTermination are all defined by me as the app supplier. I'm not expecting the device to provide this information.

What I want to avoid is having my application deployed to a device that already has someone else's app deployed using "hello-world.\<hostname\>". So, it would be ideal if either the WFM or the device could detect this if it's the case and handle it in such a way that the person installing my application can be informed that "hello-world.<hostname>" has already been used, so they can potentially configure a different prefix from "hello-world". 

As an app supplier, I could potentially handle this in a few different ways. One way would be for this prefix to be a string parameter that defaults to "hello-world," but lets the customer change it to anything they want. Or, I could potentially have an array of potential prefixes they could choose from. What I wouldn't want is for the device to pick some random prefix that the customer has no clue about.

phil-abb:
@singhmj-1 I feel like this SUP is trying to solve two different problems with the same solution, but I'm not sure the same solution should be used for both problems. I see the problems pointed out in the [issues referred to in the SUP](https://github.com/margo/specification/issues/75) as different from how we handle informing the WFM of these exclusive, limited shared, and unlimited shared capabilities. 

I see what's been started as having the potential to solve the capabilities concerns, but I'm not sure about the networking concerns.

singhmj-1:
@phil-abb  @jachstet-sea  @nilanjan-samajdar  I've pushed a revamped content of this SUP. I would appreciate some reviews and feedback. Sorry for the complexity of the content! :p

phil-abb:
> @phil-abb @jachstet-sea @nilanjan-samajdar I've pushed a revamped content of this SUP. I would appreciate some reviews and feedback. Sorry for the complexity of the content! :p

I've taken a first pass, and it's got me thinking. I'll follow up with some comments and questions today/tomorrow.

phil-abb:
> @phil-abb @jachstet-sea @nilanjan-samajdar I've pushed a revamped content of this SUP.

@singhmj-1 - Can you clarify your intention with these three parts? Since you split them up, it's presented as three seperate SUPs. Is your intention to have each part voted on (so three seperate votes), or only have one vote that covers all three parts?

singhmj-1:
@phil-abb I proposed this as a single SUP and distributed the content into 3 files for ease of understanding.

nilanjan-samajdar:
Attn: @hsinghcg 
@singhmj-1 
I am reading through the 3 parts, there's a lot to unpack.
I will add my comments over the weekend.

proposals/providers-based-resource-resolution/part1-providers.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Owner
+
+@singhmj-1
+
+---
+
+## Summary
+
+This SUP introduces a **Provider framework**: a typed, URI-identified capability contract model that gives the platform a structured language for expressing what a device offers, what is currently allocated, what an application needs from the platform, and how conflicts should be resolved — before a deployment takes palce on device.
```
phil-abb: More on this later, but this proposal is for more than just "capabilities."

proposals/providers-based-resource-resolution/part1-providers.md
```diff
@@ -0,0 +1,230 @@
...
+This SUP created a framework to solve `margo/specification#75 — Address problems exposing application endpoints` but later realized that it solves couple of additional scenarios as well.
+
+The existing specification has these gaps:
+
+1.**No declaration of resource intent.** Applications have no Margo way to 
+  declare what exclusive capabilities they will acquire on a device, so 
+  neither the WFM nor the Device can anticipate or prevent conflicts 
+  before deployment begins.
+
+2.**No dynamic value resolution.** Applications cannot receive
```
phil-abb: Not currently, but this is the same problem the [Device specific runtimes SUP](https://github.com/margo/specification-enhancements/pull/55) is trying to solve.

proposals/providers-based-resource-resolution/part1-providers.md
```diff
@@ -0,0 +1,230 @@
...
+> **⚠️ Illustrative Schemas — Not Normative**
+>
+> The provider examples(for eg, capability.margo.org/network/port, capability.margo.org/hardware/gpu) presented in this document are **illustrative only**. They exist to make the `Provider` concept concrete and reviewable. Their field names, strategies, constraints, and output structures are starting points for discussion, not normative definitions. Each provider must be formally defined via a SUP, and hence will have their own schema, with full use case documentation, TWG review, and conformance requirements.
+
+### What a Provider Is
+
+A **Provider** is a typed, named capability contract. It is not a service, a container, or a runtime component. It has no lifecycle of its own. A Provider defines:
+
+1. **A contract** — a formal schema specifying input a consumer may provide, output values the platform will resolve, and failure codes that may be returned.
+2. **An owner** — a specific scope (device or fleet) responsible for resolving the contract.
```
phil-abb: After reading through this, the "fleet" concept is unclear to me. What exactly is the intention of the fleet owner? What types of things would be considered fleet versus device?

proposals/providers-based-resource-resolution/part1-providers.md
```diff
@@ -0,0 +1,230 @@
...
+**Official Margo providers** use the `margo.org` namespace and are ratified by the TWG:
+_NOTE: These are just examples._
+
+```
+capability.margo.org/network/port
+capability.margo.org/network/ingress
+capability.margo.org/storage/volume
+capability.margo.org/hardware/gpu
+capability.margo.org/hardware/camera
+```
```
phil-abb: None of these are capabilities

>capability.margo.org/network/port

This is bindings/infrastructure assignment

>capability.margo.org/network/ingress

This is routing/network configuration

>capability.margo.org/storage/volume

This is provisioning

>capability.margo.org/hardware/gpu

This is hardware. A capability a GPU provides would be AI inference or AI training

>capability.margo.org/hardware/camera

This is hardware. A capability a camera provides would be video capture


The concepts this SUP proposes apply to several things. It's not just "capabilities" because several of your examples have nothing to do with capabilities.

proposals/providers-based-resource-resolution/part1-providers.md
```diff
@@ -0,0 +1,230 @@
...
+**Device-scope** providers are resolved by the Device Agent. The WFM MAY use cached device capability state for proactive pre-dispatch validation, but the Device Agent is the authoritative resolver. The WFM MUST NOT poll the device to query provider state.
+
+**Fleet-scope** providers are resolved by the WFM. Resolved values are injected into the `ApplicationDeployment` before dispatch. The Device Agent receives no unresolved fleet-scope bindings.
+
+### Provider Schema Structure
+
+Each provider is formally defined by a `ProviderSchema` document. Individual provider SUPs will publish these documents.
+
+```yaml
+apiVersion: margo.org/v1alpha1
```
phil-abb: I think we should keep what is required to be defined by all providers as small as possible, and allow for each provider to include any additional properties it needs. For example, you're showing "minimum", "maximum," and "pattern". I don't think the schema we have should include those. Your examples below don't use them. What should be required is just the bare minimum, and the schema should allow the provider to add any additional properties it needs.

proposals/providers-based-resource-resolution/part1-providers.md
```diff
@@ -0,0 +1,230 @@
...
+
+### Provider Scopes
+
+**Device-scope** providers are resolved by the Device Agent. The WFM MAY use cached device capability state for proactive pre-dispatch validation, but the Device Agent is the authoritative resolver. The WFM MUST NOT poll the device to query provider state.
+
+**Fleet-scope** providers are resolved by the WFM. Resolved values are injected into the `ApplicationDeployment` before dispatch. The Device Agent receives no unresolved fleet-scope bindings.
+
+### Provider Schema Structure
+
+Each provider is formally defined by a `ProviderSchema` document. Individual provider SUPs will publish these documents.
```
phil-abb: There is no explanation of what these things (request, fallback strategy, output, failure codes) are on this page.

proposals/providers-based-resource-resolution/part2-device-capabilities.md
```diff
@@ -0,0 +1,195 @@
+[<- Home](part1-providers.md#top)
+# Specification Update Proposal — Providers-Based Resource Resolution - Part 2 : Device Capabilities
+
+### Device Capabilities Extension
+We are extending the existing `DeviceCapabilitiesManifest` with an additive `providers` field that enables devices to advertise dynamic provider state to the WFM. It defines the advertisement structure, the update lifecycle. No new API endpoints are introduced — this SUP uses the existing `POST/PUT /api/v1/clients/{clientId}/capabilities` endpoint.
```
phil-abb: Keep in mind this endpoint will be changing when the Gateway spec changes are merged in: https://github.com/margo/specification/pull/180

proposals/providers-based-resource-resolution/part2-device-capabilities.md
```diff
@@ -0,0 +1,195 @@
...
+
+Without this advertisement mechanism, the WFM cannot perform pre-dispatch validation or delegate the validation to device, and the Provider concept has no operational foundation.
+
+
+### Extended `DeviceCapabilitiesManifest` Schema
+
+The existing manifest is extended with one new top-level field: `providers`. All existing fields remain unchanged.
```
phil-abb: Keep in mind the SUP to remove the rules is changing the device capabilities format a bit: https://github.com/margo/specification-enhancements/pull/50

This SUP was voted on and seems to have enough votes to be approved.

proposals/providers-based-resource-resolution/part2-device-capabilities.md
```diff
@@ -0,0 +1,195 @@
...
+### `ProviderAdvertisement` Structure
+
+Each entry in the `providers` array is a `ProviderAdvertisement` object:
+
+| Field | Type | Required? | Description |
+|---|---|---|---|
+| `provider` | string | Y | The provider URI as defined in SUP 1. Must follow the `capability.<namespace>/<category>/<name>` convention. |
+| `supported` | boolean | Y | Whether this device supports the provider. If `false`, the device is declaring awareness of the provider type but inability to fulfill it. The WFM MUST NOT dispatch deployments requiring this provider to this device. |
+| `schemaURL` | string | N | Required for custom (non-`margo.org`) providers. URL where the WFM can fetch the `ProviderSchema` document. The WFM MUST fetch and cache this schema before validating deployments against it, or this can be side-loaded into the WFM. |
+| `constraints` | object | N | Provider-specific static constraints describing what this device can offer. Structure defined by the individual provider SUP. |
```
phil-abb: I'm not following how this is connected to the provider document. It's not really explained. It's not clear to me why a device would put some information in the provider document and then a completely different set of information in the device capabilities. Are the "constraints" supposed to match what is defined as "requests" in the provider document? I can see the point of the current state to report what is allocated, but I'm not drawing the connection with the constraints and how this connects to the provider document.

proposals/providers-based-resource-resolution/part2-device-capabilities.md
```diff
@@ -0,0 +1,195 @@
...
+| `currentState` | object | N | Provider-specific dynamic allocation state describing what is currently allocated. Structure defined by the individual provider SUP. |
+
+**Note:** I was planning to propose another field like `reportInterval` in `ProviderAdvertisment` so that synchronization time of a capability can be controlled as per need. But it has been escaped due to the increasing complexity concerns.
```
phil-abb: I don't think this would make sense in this document since this is the document that would be getting sent to the WFM. If this needs to be configurable, the configuration would be somewhere else to indicate how often the device capabilities should be set.

Right now, the spec indicates it should be sent:

>To ensure the WFM is kept up to date, the device's client MUST send updated capabilities information if any changes occur to the information originally provided (i.e., additional memory is added to the device).

This makes sense to me; it should only be sent when it detects a change or after events like a reboot or something. It should be sent when nothing changes.

proposals/providers-based-resource-resolution/part3-application-deployment.md
```diff
@@ -0,0 +1,539 @@
...
+>
+> The `ApplicationDeployment` examples in this document are illustrative. The specific input and output field names within each provider's `request` block and `valueFrom` references will be defined by individual provider SUPs.
+
+---
+
+### The Key Architectural Decision
+
+Bindings live in `ApplicationDeployment` only.
+
+`ApplicationDescription` is a vendor artifact. It must be environment-agnostic. An app vendor cannot know at authoring time which port will be safe on a given device, which ingress controller is installed, or which GPU is available. Putting environment-specific binding declarations in `ApplicationDescription` would break portability — the same app could not be deployed to different devices without modifying the vendor artifact.
```
phil-abb: >Putting environment-specific binding declarations in `ApplicationDescription` would break portability — the same app could not be deployed to different devices without modifying the vendor artifact.

This is information you would actually put in the Helm chart manifests or Compose files. What we're trying to figure out is how to get that information updated based on the information the device can provide.

proposals/providers-based-resource-resolution/part3-application-deployment.md
```diff
@@ -0,0 +1,539 @@
...
+
+`ApplicationDeployment` is an operator artifact. It is created per deployment, per device. It is the correct place for environment-specific declarations.
```
phil-abb: I'm going to challenge this a bit. Why do we want the device to send information up to the WFM just so the WFM can send it back? The device knows the information, so why are we involving the WFM?

proposals/providers-based-resource-resolution/part3-application-deployment.md
```diff
@@ -0,0 +1,539 @@
...
+    ↓ reports status via Deployment Status API
+```
+
+---
+
+## Technical Proposal
+
+### `ApplicationDescription` — No Changes
+
+The existing parameter structure is unchanged. Parameters declare symbolic names and targets only:
```
phil-abb: I don't understand how the parameters are getting resolved in your examples.

How does the WFM come to this conclusion?

```YMAL
  parameters:
    - name: hostPort
      valueFrom: webPort.result.assignedPort
      targets:
        - pointer: service.nodePort
          components:
            - hello-world
```

You cannot assume just because the parameter name has "port" in it, the app wants the WFM to assign a port number to it. 

This is why I think it makes sense to incorporate the ideas expressed in the [device-specific parameters SUP](https://github.com/margo/specification-enhancements/pull/55).

The provider document can indicate what the parameter value key is. The app description can indicate that it wants the parameter's value to come from the device using the key. When the device gets the application deployment with the indication of the device-specific parameter key to replace the value for, it does its magic to replace the expression with whatever port number it wants to use. 

To me, it feels like we are mixing concepts. I think it makes sense for the WFM to have the information it needs to do some pre-matching and maybe some level of pre-validation, but not be responsible for assigning these values.

proposals/providers-based-resource-resolution/part3-application-deployment.md
```diff
@@ -0,0 +1,539 @@
...
+        components: ["my-app-compose"]
+```
+
+The app vendor ships one `ApplicationDescription`. It works on any Margo-compliant device regardless of what port is actually assigned or which GPU device path is used. The binding resolution is entirely the operator's and platform's concern.
+
+### `ApplicationDeployment` Extensions
+
+There are two changes to the `spec` section of `ApplicationDeployment`:
+
+#### `requiredCapabilities`
```
phil-abb: I think I have a fundamental misunderstanding of how you intended the provider document to be used. I don't understand why the WFM would tell the device information it already knows. The device is the one that knows what it can provide, not the WFM.

proposals/providers-based-resource-resolution/part3-application-deployment.md
```diff
@@ -0,0 +1,539 @@
...
+```
+
+### WFM Pre-Dispatch Validation Flow
+
+Before dispatching an `ApplicationDeployment` to a device, the WFM MUST execute the following validation steps in order. If any step fails, the WFM MUST reject the deployment and return the corresponding error. The deployment MUST NOT be dispatched to the device.
+
+```
+Step 1: For each requiredCapabilities entry, validate provider scope
+
+  Is the scope neither "fleet" nor "device"?
```
phil-abb: The SUP doesn't explain this fleet concept very well. I cannot tell from what is provided so far what this really means.

proposals/providers-based-resource-resolution/part1-providers.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Technical Proposal
+
+> **⚠️ Illustrative Schemas — Not Normative**
+>
+> The provider examples(for eg, capability.margo.org/network/port, capability.margo.org/hardware/gpu) presented in this document are **illustrative only**. They exist to make the `Provider` concept concrete and reviewable. Their field names, strategies, constraints, and output structures are starting points for discussion, not normative definitions. Each provider must be formally defined via a SUP, and hence will have their own schema, with full use case documentation, TWG review, and conformance requirements.
+
+### What a Provider Is
+
+A **Provider** is a typed, named capability contract. It is not a service, a container, or a runtime component. It has no lifecycle of its own. A Provider defines:
```
phil-abb: After going through and reading/commenting on all three parts of this SUP, it's clear to my I really don't understand how you intend for these provider documents to be used. I thought I had an understanding, but once I got to part 3, it's clear the way I thought they would be used is not what you intended. 

Part 1 needs to be updated, with more information about what this provider document is for, how it would be created, how it's used by the WFM and device, and how they will be "cataloged" by Margo so suppliers know what needs to be provided and how to use what is provided.

phil-abb:
To add to my comments about using the term "capabilities'.

First, I think we have chosen poorly when we called the document "DeviceCapabilities". I think this is leading in a bad direction, where we're calling everything a "capability" even when it's not. There are clearly things in this device capabilities document that have nothing to do with capabilities (e.g., vendor, model number, serial number, etc.). This isn't something I expect this SUP to solve, but I think it's caused us to use the term "capability" in a very generic way to talk about things as capabilities that aren't (e.g, network configuration, provisioning, bindings, equipment, hardware, etc), which I commented on in this SUP.

From how I'm understanding this SUP, it's actually attempting to solve several different types of problems, and capabilities identification is just one of those problems.

One of the examples I pointed out. I've seen a few different references to a camera being a capability. The camera is a piece of equipment that has the capability to capture video. The camera itself isn't a capability.

I did some AI-assisted research, and I think we need to think a little bit differently about how we are using and referring to capabilities. This is a table we came up with that shows what I mean by identifying several capabilities Margo may be interested in handling, how those resources are typically allocated (Shared = unlimited allocations, Limited Shared = X allocations allowed, Exclusive = 1 allocation allowed), and common properties and applications suppliers may need to indicate their app needs to be compatible.

| Category | Capability Name | Sharing Model | Key Properties Apps Will Match On |
|---|---|---|---|
| Media / Video | video.capture | Limited shared | resolution, frameRate, codec |
| Media / Video | video.encode | Limited shared | codec, maxResolution |
| Media / Video | video.decode | Limited shared | codec, maxResolution |
| AI / Compute Acceleration | ai.inference | Limited shared | runtime (tflite, onnx, cuda), modelType, performance (TOPS) |
| AI / Compute Acceleration | ai.training | Limited shared | acceleratorType, memory |
| General Compute (specialized) | compute.gpu | Limited shared | memory, vendor (optional), architecture |
| General Compute (specialized) | compute.accelerator | Limited shared | type (tpu, vpu, fpga), runtime |
| Industrial / IoT Protocols | industrial.opcua | Limited shared | securityMode, endpointFeatures |
| Industrial / IoT Protocols | industrial.modbus | Limited shared | protocolVariant (tcp, rtu), baudRate |
| Industrial / IoT Protocols | industrial.fieldbus | Limited shared | protocol (can, profibus, etc.) |
| Interfaces (logical, not hardware-level) | serial.interface | Exclusive or limited shared | protocol, baudRate |
| Interfaces (logical, not hardware-level) | gpio.interface | Exclusive | pinCount, capabilities (input/output/pwm) |
| Interfaces (logical, not hardware-level) | i2c.interface | Limited shared | busSpeed |
| Networking | network.connectivity | Unlimited or limited shared | type (ethernet, wifi), bandwidth |
| Networking | network.5g | Limited shared | carrier, bandwidth, latency |
| Networking | network.lowLatency | Limited shared | maxLatency |
| Storage / Data Path | storage.nvme | Limited shared | iops, throughput, latency |
| Storage / Data Path | storage.local | Limited shared | capacity, accessSpeed |
| Security / Trust Hardware | security.tpm | Limited shared | version |
| Security / Trust Hardware | security.hsm | Limited shared | operationsSupported, throughput |
| Sensors | sensor.camera (alias of video.capture in some models) | Limited shared | resolution, fieldOfView |
| Sensors | sensor.temperature | Unlimited or limited shared | accuracy, range |
| Sensors | sensor.lidar | Limited shared | range, resolution |
| Actuation / Robotics | robotics.actuator | Exclusive | type (arm, motor), degreesOfFreedom |
| Actuation / Robotics | robotics.controller | Exclusive | supportedDevices |
| Audio | audio.capture | Limited shared | sampleRate, channels |
| Audio | audio.processing | Limited shared | features (noiseSuppression, etc.) |
| Position / Location | location.gps | Limited shared | accuracy, updateRate |
| Time / Sync | time.sync | Unlimited shared | precision (ms, µs) |
| Power / Energy (edge-specific) | power.capacity | Limited shared | wattsAvailable |
| Peripheral Abstraction (fallback) | peripheral.generic | Exclusive or limited shared | type (broad classification only) |

phil-abb:
I made a couple of comments about seeing this SUP as solving multiple problems and not just a "capability" problem. 

In my [comment above](https://github.com/margo/specification-enhancements/pull/66#issuecomment-4602849425), I indicated:
> I feel like this SUP is trying to solve two different problems with the same solution, but I'm not sure the same solution should be used for both problems.

But after reading through this again, I think it might actually work once the idea is refined. 

In more general terms, I see this SUP looking to provide a solution for the following problems:

### I want to use the parameter value "X" in my manifest. Is this ok?

This is something like the ingress route configuration problem. I want to be able to ask the device if it's ok to use "hello-world" for my ingress route prefix so my ingress manifests `host` property becomes `hello-world.<device hostname>`. The device can either indicate yes, it's ok, or no, someone already has an ingress route configured to `hello-word.<device hostname>`. I don't want the device to assign some random value to it because the customer needs to actually know how to get to the hello-world portal, and I don't want the URL to be some random string. Instead, I want to give the customer installing the app a chance to use a different prefix like `hi-world`.

This has nothing to do with capabilities

### I want the device to give me an "X" value for my parameter to use in my manifest

This is an example of the port assignment problem. As an application supplier, I may not be too concerned about what port number is used, so I'm just asking the device to assign me one.

There may be variations of this where the app supplier does want the person installing the app to know the port number, so this would fall more into the first category.

This also has nothing to do with capabilities

### I want to use the parameter value "X" in my manifest if I can. If not, I want the device to give me a value I can use.

This is a combination of the first two examples

This is a case where maybe you would want to give the customer a chance to use a specific value if they can, but it's not important enough that the device shouldn't assign a different value to it. The most obvious case of this is if the app description defines a default value, but there are probably other examples of wanting this.

This also has nothing to do with capabilities.

### I want to know if device "X" is able to run my app that needs the capability "Y."

I have an app that monitors for safety issues. It's expecting a device that has the capability to do video capture. This is a match-making problem, so the application description needs to be able to indicate it requires video capture, a device needs to be able to indicate it can do video capture, and the WFM can do the match-making if it wishes.

This does deal with capabilities

### I need the device to allocate "X" to me so that I can do "Y".

I have an app with the capability to do AI Inference. It needs to be deployed to a device that has the capability of doing AI inference, and that device needs to be able to allocate 1 unit of the N available GPUs. If the device has at least 1 unit free to allocate, the application should be deployable there. If it doesn't, the app probably won't be deployable there. 

**Note:** This isn't saying Margo should be responsible for doing that allocation (see below)

This does deal with capabilities

### I have found a device that can allocate "X" to me so I can do "Y". I need the device to provide the information for my parameter so the Helm or Compose manifests can be updated to request it.

(Strong opinion) It shouldn't be Margo's responsibility to actually do any allocations of anything. This should be the responsibility of the Kubernetes scheduler or Compose, and not something Margo does. Margo should just be responsible for communicating what is needed, what is available, and how to configure the manifest to request what it needs from Kubernetes or Compose.

Continuing with the AI Inference above. My Kubernetes application needs to be deployed to a device that can do AI inference. My application is built so it can run on devices with different GPUs, so I'm not too concerned about matching vendor NVIDIA or AMD, for example. Since I don't know what GPU the device is going to have, I cannot include the following in my pod manifest

```yaml
   resources:
      limits:
        nvidia.com/gpu: 1
```

I need the device to tell my app, via a parameter value, if it's "nvidia.com/gpu" or "amd.com/gpu" so that when my Kubernetes application is deployed, that resource is actually allocated by Kubernetes.

This is an easy example because there are device plugins for doing this, and it's a simple string that is needed. For Compose and other types of allocations, it's probably more difficult.

This is about capabilities and resource allocations

proposals/resource-allocation-to-workloads.md
phil-abb: The updates to the SUP included some helpful use cases.

proposals/providers-based-resource-resolution/part1-providers.md
singhmj-1: If we limit provider scopes strictly to the local 'device,' we unintentionally push WFM vendors into a corner and force them to build messy workarounds if they want to resolve any resource centrally. Hence, the `fleet` scope is introduced. I'm writing down some imaginary scenarios that come to my mind:

**AWS Secrets Manager Injection:** An app needs an AWS key to upload data safely. Instead of hardcoding keys or manually creating 500 different passwords for 500 devices, the developer uses a fleet provider. The WFM intercepts the deployment, fetches a secure, short-lived key from AWS Secrets Manager, and bakes it into the app's configuration. The application runs on device perfectly without ever knowing the secrets manager exists.

**Licensing and Feature Toggling:** An app unlocks its features based on the licensing tier bought by a customer. During deployment, the WFM checks the customer account and resolves the license in the configuration before making it available to the device.

proposals/providers-based-resource-resolution/part1-providers.md
singhmj-1: I agree with you and think that we should call them a `resource` rather than a `capability`.
```
resource.margo.org/network/port
resoure.margo.org/network/ingress
resource.margo.org/storage/volume
resource.margo.org/hardware/gpu
resource.margo.org/hardware/camera
```

proposals/providers-based-resource-resolution/part1-providers.md
singhmj-1: ### Provider Schema Structure

As already mentioned, each provider is formally defined by a `ProviderSchema` document. A `ProviderSchema` is the **immutable public contract** for a provider with some inputs, and outputs. It has four parts:

| Part | What it defines |
|---|---|
| `request` | The input fields a provider contract supports. Think of it as the API request body. |
| `fallbackStrategies` | The exhaustive list of strategies the provider supports to execute if the original request cannot be satisfied. The provider defines in its contracts what strategies it supports and what individual strategy mean. |
| `output` | The fields the provider-contract execution will return upon successful resolution. |
| `failureCodes` | The explicit error constants the provider will return if resolution fails unrecoverably. Each code is machine-readable, defined and documented in the contract itself. |

Hope this makes sense. I'll add this block to the docs, if not further clarification is needed.

proposals/providers-based-resource-resolution/part1-providers.md
singhmj-1: I agree that it will provide extensibility. I'll brainstorm this part again.

proposals/providers-based-resource-resolution/part2-device-capabilities.md
singhmj-1: I'll update this once Gateway spec is merged and reflected on the official docs. Thanks!

proposals/providers-based-resource-resolution/part2-device-capabilities.md
singhmj-1: Thanks, will update this once #50 is merged.

proposals/providers-based-resource-resolution/part2-device-capabilities.md
singhmj-1: **ProviderSchema** is consumed by application operators — when authoring an ApplicationDeployment, the operator picks the providers it would need and read the schema to know what request fields they can send and what output fields they can receive so that they can be wired into parameters.

**ProviderAdvertisement** is offered by the device — it declares which providers it supports, what resources it has available (constraints), and what is already taken (currentState).

The `constraints` in `ProviderAdvertisement` is not derived from the `ProviderSchema` output fields — it is the device's own description of its resources in a structure way.

proposals/providers-based-resource-resolution/part2-device-capabilities.md
singhmj-1: Agreed. I'll remove this note in the next commit. 

That said, I want to flag one open concern: the current spec has no snapshot or periodic reconciliation mechanism, which means any drift in reported capabilities — caused by a crashed Device Agent, a lost update, or a silent resource release — has no defined recovery path. I'm not proposing we solve it in this SUP, but I think it's worth capturing as an open question so it doesn't get lost.

proposals/providers-based-resource-resolution/part1-providers.md
phil-abb: I don't think "resources" is correct either. We don't need to use the same term to refer to all of them. We should refer to them as what they are

singhmj-1: Okay then would stick to `provider.margo.org/network/port` .

proposals/providers-based-resource-resolution/part1-providers.md
phil-abb: If you want to include this, there needs to be more information included about it in your SUP. We need to have Margo-specific use cases explaining why these fleet-level providers are needed and how they are used, because we would need a Margo-curated catalog of these as well for them to be useful in an interoperable way.

proposals/providers-based-resource-resolution/part1-providers.md
phil-abb: I'm ok with this for now.

proposals/providers-based-resource-resolution/part1-providers.md
phil-abb: > As already mentioned, each provider is formally defined by a ProviderSchema document.

While this may be the case, the ProvideSchema document (Part 1) doesn't acutally explains them. As commented below, Part 1 seems to be missing a lot of key information.

proposals/providers-based-resource-resolution/part2-device-capabilities.md
phil-abb: > ProviderSchema is consumed by application operators

What are "application operators"? Unless I missed it, the SUP doesn't explain this. It mentions operators, but it's unclear what the SUP means by "operators" or what they actually do. Currently, the WFM is responsible for creating the Application Deployment artifacts, so it's unclear what changes you're proposing. Also, what do you mean by "application" here? 

> The constraints in ProviderAdvertisement is not derived from the ProviderSchema output fields — it is the device's own description of its resources in a structure way.

So how does the WFM know what these "constraints" are if they don't match anything in the provider?

proposals/providers-based-resource-resolution/variant-2/proposal.md
phil-abb, line 1: This update helps answer some of the questions I had. I would suggest putting more detailed examples in the proposal markdown because I didn't really understand what this was actually proposing until I looked at the use-case document.

One of the big parts that wasn't clear to me by just reading the proposal markdown was how the "stateSchema" was used. The current write-up doesn't cover the "stateSchema" at all.

singhmj-1: Thanks for reviewing, I'll be pushing some more changes today.

singhmj-1: I'm still brainstorming on this, and I feel like it can be simplified. I'll push once finished on my side and will leave a "Request for review". Thanks!

phil-abb:
A [SUP has been created](https://github.com/margo/specification-enhancements/pull/76) as a precursor to this SUP. It pulls out some of the ideas that were started here and improves on them. Once that SUP develops more, this SUP will be updated as a follow-up to it.

singhmj-1:
I have updated the content of this SUP post extracting the https://github.com/margo/specification-enhancements/pull/76 SUP from it.

I think that there is still some work needed in this SUP, hence would appreciate some feedback.

phil-abb:
> I think that there is still some work needed in this SUP, hence would appreciate some feedback.

@singhmj-1 - I'll try to take a look later this week.

proposals/profile-discovery-and-resolution/01-capability-framework.md
```diff
@@ -0,0 +1,596 @@
...
+
+There is no standard mechanism for applications to declare "I need a GPU with 8 GiB VRAM" or "I need an available network port" or "I need a secret from the vault." Each implementation invents its own approach. No cross-implementation conflict detection is possible.
+
+---
+
+## Framework Concepts
+
+### Extending ProfileDefinition
+
+As per [`sup_bisect_device_manifest`] :
```
phil-abb, line 68: Did you intend for the "[`sup_bisect_device_manifest`]" part to be a link to the other SUP?

proposals/profile-discovery-and-resolution/01-capability-framework.md
```diff
@@ -0,0 +1,596 @@
...
+  "NoGPUAvailable",
+  "InsufficientVRAM"
+]
+```
+
+## Discovery vs. Allocation Policy
+
+**Important distinction:** Discovery is inherently a declarative inquiry, not a guarantee of reservation. The framework allows an application to declare its resource requirements; the platform then evaluates this declaration against the live infrastructure state.
+
+The core specification defines only the resolution contract (inputs, outputs, and failure modes). It deliberately decouples resource resolution from lifecycle enforcement, and how the resolution is done are implementation defined, for example:
```
phil-abb:
```suggestion
The core specification defines only the resolution contract (inputs, outputs, and failure modes). It deliberately decouples resource resolution from lifecycle enforcement, and how the resolution is done is an implementation detail, for example.
```

proposals/profile-discovery-and-resolution/01-capability-framework.md
```diff
@@ -0,0 +1,596 @@
...
+
+* Transient Resolution: A storage class discovery may simply return a metadata name without locking the underlying pool.
+* Implicit Reservation: A network port discovery may mark the resolved port as occupied immediately upon evaluation.
+* Exclusive Enforcement: A GPU discovery may trigger strict hardware isolation or exclusive runtime access.
+
+Because resource types have fundamentally different lifecycles, the exact enforcement policy is deferred to the underlying platform implementation.
+
+Note: Need some feedback/guidance/brainstorming on whether the resource resolution should be defined by Margo, otherwise there will not be a deterministic behaviour.
```
phil-abb, line 209: What, specifically, do you mean by "resource resolution"? 

Based on the text above this note, I'm interpreting this as how the client handles allocating the resources (if they are allocatable). I don't think Margo should be defining any implementation details. The way a device supplier chooses to implement these things is inside the black box. Margo should focus on how the required information is communicated to all parties needing the information. 

However, since you mentioned not defining this makes things non-deterministic, maybe I'm misinterpreting.

proposals/profile-discovery-and-resolution/01-capability-framework.md
```diff
@@ -0,0 +1,596 @@
...
+5. **Execution time** — Deployment executes with concrete values
+
+### Atomicity
+
+Resolution of all `discoverProfiles` entries within a single `ApplicationDeployment` is **atomic**: either all entries resolve successfully or the entire deployment is rejected. Partial resolution is not permitted. If any single `ProfileDiscoveryRequest` fails, the platform MUST:
+
+- Reject the deployment before any component is installed
+- Return a structured failure identifying which `profileDefinitionId` caused the failure and the applicable failure code
+
+This aligns with the all-or-nothing batch semantics established in [`sup_bisect_device_manifest`] for `ProfileState` submission.
```
phil-abb, line 315: Was this "[`sup_bisect_device_manifest`]" part intended to be a link as well?

proposals/profile-discovery-and-resolution/01-capability-framework.md
```diff
@@ -0,0 +1,596 @@
...
+- Introducing fleet scope now would create normative obligations on WFM vendors before the contract is sufficiently specified
+- Device-scoped resolution is the foundational building block; fleet scope builds on top of it, and then other options can also be.
+
+---
+
+## Use Cases
+
+### Use Case 1 — GPU Discovery (Generator)
+
+An ML inference application needs a GPU with at least 8 GiB VRAM and Ampere architecture.
```
phil-abb, line 424: This use case is missing the ApplicationDescription YAML showing how the app vendor indicates it needs an 8GiB VRAM GPU with Ampere. 

It would be helpful if each use case showed the ApplicationDescription,yaml (can be snippets) the app vendor provides, the ApplicationDeployment.yaml the WFM provides, and the ProfileDefinition.

proposals/profile-discovery-and-resolution/01-capability-framework.md
```diff
@@ -0,0 +1,596 @@
...
+
+**ApplicationDeployment:**
+```yaml
+discoverProfiles:
+  servicePort:
+    id: deviceprofile.margo.org/network/dynamic-port
+    request:
+      protocol: tcp
+      rangeMin: 8000
+      rangeMax: 9000
```
phil-abb, line 556: This is not information the WFM needs to tell the device because the device already knows this. 

The device sends its `deviceprofile.margo.org/network/dynamic-port` to the WFM, so it doesn't need to send it back to the device.

proposals/profile-discovery-and-resolution/01-capability-framework.md
```diff
@@ -0,0 +1,596 @@
...
+```
+
+**ApplicationDeployment:**
+```yaml
+discoverProfiles:
+  httpPort:
+    id: deviceprofile.margo.org/network/port
+    request:
+      port: 8080
+      protocol: tcp
```
phil-abb, line 509: This is kind of confusing. I think I need to see the ApplicationDescription.yaml to understand what is happening. The fact that an app wants to use TCP on Port 8080 needs to come from the ApplicationDescription, so I'd like to see how you envision this information being provided.

proposals/profile-discovery-and-resolution/01-capability-framework.md
```diff
@@ -0,0 +1,596 @@
...
+
+Without `mode`, the platform must infer evaluation behaviour from schema shape — an unreliable heuristic. Explicit `mode` declaration makes platform behaviour deterministic and removes ambiguity for implementors.
+
+---
+
+## Open Questions
+
+| # | Question | Status |
+|---|---|---|
+| 1 | Should `requestSchema` support `required` fields, or are all request inputs always optional? | Open |
```
phil-abb, line 594: Since each device profile has its own schema definition, this can be determined independently.

proposals/profile-discovery-and-resolution/01-capability-framework.md
```diff
@@ -0,0 +1,596 @@
...
+| 2 | Should `failureCodes` support structured metadata (e.g., human-readable descriptions) in addition to machine-readable codes? | Open |
+| 3 | Should `valueFrom` support a default value fallback if resolution produces no output for an optional field? | Open |
```
phil-abb, line 596: I think it should be possible to define a default value, but the behavior is the opposite. The default value is used first if it can be used. For example, I may want the port to default to 1234 if it's available; otherwise, another port is assigned.

proposals/profile-discovery-and-resolution/01-capability-framework.md
```diff
@@ -0,0 +1,596 @@
...
+
+### Problem 3 — Discovery logic is scattered and inconsistent
+
+There is no standard mechanism for applications to declare "I need a GPU with 8 GiB VRAM" or "I need an available network port" or "I need a secret from the vault." Each implementation invents its own approach. No cross-implementation conflict detection is possible.
+
+---
+
+## Framework Concepts
+
+### Extending ProfileDefinition
```
phil-abb, line 66: I think something may have been lost in the re-write/edits. In the original idea, there was a mechanism for the device to indicate how many allocatable units were available (e.g., GPUs) so the WFM could determine if the device had that resource free. 

Between the rewrite of this SUP and the creation of the other SUP, I'm not seeing that anymore.

proposals/profile-discovery-and-resolution/01-capability-framework.md
```diff
@@ -0,0 +1,596 @@
...
+
+Because resource types have fundamentally different lifecycles, the exact enforcement policy is deferred to the underlying platform implementation.
+
+Note: Need some feedback/guidance/brainstorming on whether the resource resolution should be defined by Margo, otherwise there will not be a deterministic behaviour.
+
+---
+
+### ProfileDiscoveryRequest — A Materialization of the newly introduced discoverySchema
+
+A **ProfileDiscoveryRequest** is authored by an end-user persona or wfm inside an `ApplicationDeployment`. It is a materialization of the discovery schema defined in the `ProfileDefinition`. It is a **deferred declaration** — authored before deployment, resolved by the platform at evaluation time (when the deployment is triggered).
```
phil-abb, line 215: The WFM is the only thing that will create the ApplicationDeployment.yaml

```suggestion
A **ProfileDiscoveryRequest** is authored by the wfm inside an `ApplicationDeployment`. It is a materialization of the discovery schema defined in the `ProfileDefinition`. It is a **deferred declaration** — authored before deployment, resolved by the platform at evaluation time (when the deployment is triggered).
```

proposals/profile-discovery-and-resolution/01-capability-framework.md
```diff
@@ -0,0 +1,596 @@
...
+* Implicit Reservation: A network port discovery may mark the resolved port as occupied immediately upon evaluation.
+* Exclusive Enforcement: A GPU discovery may trigger strict hardware isolation or exclusive runtime access.
+
+Because resource types have fundamentally different lifecycles, the exact enforcement policy is deferred to the underlying platform implementation.
+
+Note: Need some feedback/guidance/brainstorming on whether the resource resolution should be defined by Margo, otherwise there will not be a deterministic behaviour.
+
+---
+
+### ProfileDiscoveryRequest — A Materialization of the newly introduced discoverySchema
```
phil-abb, line 213: After reading through the updates, I'm confused by this `discoverProfiles` object. The SUP doesn't fully explain how the information in this section gets from the ApplicationDescription. Also, in some of the examples below, it seems like it's just a repeat of what the device sent to the WFM when it sends up its device profile object.

proposals/profile-discovery-and-resolution/02-extensibility.md
```diff
@@ -0,0 +1,211 @@
...
+      {
+        "partitionId": "fpga-partition-0",
+        "bitstreamSlots": 4
+      }
+    ]
+  }
+}
+```
+
+**Step 3 — Application operator declares a `ProfileDiscoveryRequest`:**
```
phil-abb, line 182: What persona is the "Application operator?" 

- Is the application supplier? If so, then I'd suggest using "Application supplier" or "Application vendor"
- Is this the on-site person who is clicking the button to install the app? If so, I'd suggest "WFM supplier" or "WFM vendor" since it's the entity that created the ApplicationDeployment.yaml

proposals/profile-discovery-and-resolution/02-extensibility.md
```diff
@@ -0,0 +1,211 @@
+# Extensibility
```
phil-abb, line 1: I think this page would be good to move to the [other SUP](https://github.com/margo/specification-enhancements/blob/sup/bisect-device-manifest/proposals/sup_bisect_device_manifest.md) as an addendum. I don't think this is anything to vote on specifically, but it highlights how what you're proposing in the other SUP can be used to accomplish this evolution.

proposals/profile-discovery-and-resolution/01-capability-framework.md
```diff
@@ -0,0 +1,596 @@
...
+Without `mode`, the platform must infer evaluation behaviour from schema shape — an unreliable heuristic. Explicit `mode` declaration makes platform behaviour deterministic and removes ambiguity for implementors.
+
+---
+
+## Open Questions
+
+| # | Question | Status |
+|---|---|---|
+| 1 | Should `requestSchema` support `required` fields, or are all request inputs always optional? | Open |
+| 2 | Should `failureCodes` support structured metadata (e.g., human-readable descriptions) in addition to machine-readable codes? | Open |
```
phil-abb, line 595: I'm ok with just the failure code; that way, the WFM is free to construct its error messages in a way that is consistent with the rest of their site.

A side note on this is globalization. Right now, everything in Margo only supports a single language. At some point, we're going to have to figure out how to support multiple languages. Having human-readable messages here would just add additional overhead to be able to provide these error messages in a bunch of different languages.

proposals/profile-discovery-and-resolution/01-capability-framework.md
```diff
@@ -0,0 +1,596 @@
...
+---
+
+### Use Case 3 — Available Network Port (Generator)
+
+An application needs any available port assigned dynamically rather than specifying one explicitly.
+
+**ProfileDefinition:**
+```json
+{
+  "profileDefinitionId": "deviceprofile.margo.org/network/dynamic-port",
```
phil-abb, line 523: I'm not a fan of devices having to provide both `deviceprofile.margo.org/network/port` and `deviceprofile.margo.org/network/dynamic-port` profiles. There's too much duplication.

I think it would be better if a profile could handle both cases instead.

proposals/profile-discovery-and-resolution/01-capability-framework.md
singhmj-1: By "resource resolution," I mean the exact point where the platform evaluates the application's declared requirements against the live infrastructure state and decides it's a match.

phil-abb: This should be the device's responsibility. The WFM should just be responsible for passing the information the device requires via the application deployment YAML. The device should handle everything that's needed and report any errors it has.

singhmj-1: You are completely right that the internal implementation details of how a vendor manages its resources should remain a black box. However, the text currently doesn't discuss the the side effects of discovery and leaves them entirely up to the implementation (e.g., Transient Resolution vs. Implicit Reservation), we introduce a silent race condition that breaks deterministic behavior.

For Example:
* Imagine App 1 and App 2 are pushed to a device at the exact same time, and both require Port 8080.
* If Vendor A implements Implicit Reservation: App 1's discovery phase runs first and immediately marks Port 8080 as occupied. A millisecond later, App 2's discovery phase runs, sees it's occupied, and safely fails.  (This is deterministic behavior).
* If Vendor B implements Transient Resolution: App 1's discovery check runs and returns true without locking. Before App 1 can fully deploy and bind the port, App 2's discovery check runs, sees the port is still technically free, and also returns true. Both pass discovery, both attempt to deploy, and one crashes randomly at runtime. (Non-deterministic behavior). 

The Discovery mechanism fails here. Vendors could fix this race condition by doing a soft-allocation or hard-allocation during the discovery phase, but as the text stands, that is completely up to them. Some vendors might not want to do it.

If Margo's contract states that “Discovery is not a guarantee of reservation,” then Margo must define a standard discovery detail as well. Otherwise, an application that is perfectly compliant with the Margo spec will work flawlessly on Vendor A's hardware but crash randomly on Vendor B's hardware.

Does that map show where the non-deterministic behavior sneaks in?

phil-abb: Ah, ok. I understand what you mean now. 

Isn't this already handled, though, based on how you defined [atomicity](https://github.com/margo/specification-enhancements/blob/feat/infra-allocation-and-conflict-resolution/proposals/profile-discovery-and-resolution/01-capability-framework.md#atomicity)

With the way the desired state works, the device will pull down the latest desired state, which would indicate two applications needing to be deployed, so the device starts deploying them

- With the atomicity rule, the first app would see that port 8080 is free, so the deployment process would continue for that app. Once the deployment for that app is successful, it moves to the next app.
- With the atomicity rule, it would see that port 8080 is no longer free, so the deployment of that app would fail.

Shouldn't the act of having a successful deployment "claim" that resource if it's allocatable?

Maybe I'm making some assumptions that the spec doesn't define as far as apps being deployed sequentualy install of in parallel.

proposals/profile-discovery-and-resolution/01-capability-framework.md
singhmj-1: That mechanism had a fundamental flaw. The goal was to let WFMs run a quick pre-check against the device profile state, but that state is essentially a cached, stale entity. Given how fast local device activity changes and how often they are reported to WFM, the pre-check wouldn't be accurate.

singhmj-1: It was just an enhancement, it's safer to leave it out for now and add it later if we find a clear need.

proposals/profile-discovery-and-resolution/01-capability-framework.md
singhmj-1: Yeah. I'll push it. Thanks.

proposals/profile-discovery-and-resolution/01-capability-framework.md
singhmj-1: hmm.. that assumption can solve this problem but actually it is nowhere specified in the spec and we never considered sequential guarantee even while writing sandbox. So, I think if we want to mandate sequential installs, then we should add this to the official spec so that everyone be well aware of this.

proposals/profile-discovery-and-resolution/01-capability-framework.md
singhmj-1: I'm not sure if the spec actually has any stance on this, but I had an impression that an `ApplicationDeployment` can be authored in multiple ways and one of them could be:

> An end-user comes to a WFM Marketplace, selects an application description and overrides whatever params are allowed, and then triggers a deployment.

Now this is completely upto the WFM implementation how they prepare the `ApplicationDeployment`. It could be prepared completely on a UI and then sent to the WFM, or it can be done completely on the WFM. It can ask user to override params as well. This is why, I added `authored by end-user persona or wfm`, as I am not sure what the implementation could be.

proposals/profile-discovery-and-resolution/01-capability-framework.md
singhmj-1: The `discoverProfiles` is deliberately never a part of the `ApplicationDescription`. The reason for this is to enforce a strict separation of concerns. My take is that we should keep the developer completely out of the discovery business.

The Developer defines the "Contract": All they should do is expose clean, well-documented parameters stating exactly what the application needs to run (e.g., gpu_required: AMD or default_port: 8080). They shouldn't have to think about infrastructure discovery mechanisms at all.

The End-User/WFM handles the "Binding": It should be completely up to the end-user or the WFM to write the profile discovery. They use the it to bind those abstract needs to actual, live hardware at execution time.

If we let developers write profile discovery inside the application description, we create a massive maintenance risk: if we ever make a change to the `ProfileDefinition` in official spec then every single application description using it will go invalid overnight. 

But, by limiting this responsibility strictly to the end-user/WFM at the deployment layer (ApplicationDeployment object), the impact surface area of a spec update becomes much smaller and easier to manage.

Just my take. Pardon if it sounds too naive or assumptive.

proposals/profile-discovery-and-resolution/01-capability-framework.md
singhmj-1: Yup, I'll fix it. Thanks!

proposals/profile-discovery-and-resolution/01-capability-framework.md
phil-abb: Marketplaces are not in scope for Margo. 

I'll chalk this up to semantics, so I'll drop this comment.

proposals/profile-discovery-and-resolution/01-capability-framework.md
phil-abb: I think we're misunderstanding each other.

Let's take this requirement as an example. 

I am an application supplier with an application that requires a GPU with at least 8 GiB of RAM and Ampere architecture. I would like the WFM to be able to match my application requirements to a compatible device.

## Current Specification
In the current specification, we have a limited way for the device supplier to indicate what the device can provide, and a limited way for the application vendor to indicate what their application needs. With the information from both sides, the WFM can attempt some smart pairing of application to device.

- Device supplier [Device Capabilities](https://docs.margo.org/specification/margo-management-interface/device-capabilities) - Through the current device capabilities, the device supplier can indicate their device has a GPU by specifying a peripheral of type `gpu`. However, our current implementation is limited, and the device supplier cannot indicate how much RAM or what architecture the GPU is.
- Application supplier [Application Description](https://docs.margo.org/specification/applications/application-description#requiredresources-attributes) - Through the current application description, the application supplier can indicate their application requires a GPU by setting this in the `RequiredResources` section of the application description. However, since the current `RequiredResources` is a direct map to `Resources` in the device capabilities, the application supplier also cannot indicate how much RAM and what architecture the GPU needs.

## Device Manifest Bisection SUP
The [new SUP](https://github.com/margo/specification-enhancements/blob/sup/bisect-device-manifest/proposals/sup_bisect_device_manifest.md) you're working on re-imagines how device capabilities should be implemented, and it provides for a more dynamic and detailed way for a device to express what it has and what it can do. This SUP is scoped to the device capabilities only and does not propose any changes to the application description (I'm not saying it needs to, just using this to clarify my original point)

- Device supplier device profiles - Through the enhanced profiles SUP, the device supplier now has a way to express much more details about the GPU on their device via the `deviceprofile.margo.org/peripherals/gpu`. So, now, as a device supplier, I have a way to communicate not only that I have a GPU, but it has 24 GiB of vRAM and the "Ampere" architecture.
- Application supplier [Application Description](https://docs.margo.org/specification/applications/application-description#requiredresources-attributes) - There are currently no proposed changes to the application description, so even though the device supplier is capable of communicating that their device has a GPU with 24GiB of vRAM and the "Ampere" architecture, the application supplier is still only limited to saying they need a GPU and not how much RAM or what architecture they need

## Profie Discovery & Resolution
This [additional follow-on SUP](https://github.com/margo/specification-enhancements/blob/feat/infra-allocation-and-conflict-resolution/proposals/profile-discovery-and-resolution/01-capability-framework.md) introduces some new layers on top of the device profiles. It provides a mechanism for how an application vendor can use a combination of the functionality being proposed in the [device-specific parameter value SUP](https://github.com/margo/specification-enhancements/pulls) with the new functionality in this profile discovery SUP. These proposed changes now allow the application vendor to get parameter values that are required (e.g., the devicePath), but it still doesn't address how the application vendor indicates their application requires a GPU with at least 8 GiB of RAM and Ampere architecture

## Original question
So, my original question was intended to see if you had given any thought to how the ApplicationDescription would be updated to also allow the application supplier to indicate more details about what their application's required resources are.

- Do you have any intentions on defining changes to the ApplicationDescription in either of these SUPs, or a follow-on SUP?
- If you've thought about it, do you see it as extending the current application description's [RequiredResources](https://docs.margo.org/specification/applications/application-description#requiredresources-attributes) object to allow specifying more details about the required resources? Or do you see a more significant change also needed in the application description so things are more dynamic for the application supplier as well?
- If you haven't considered this, is this something you would expect someone else to define as a companion to your SUPs?

proposals/profile-discovery-and-resolution/01-capability-framework.md
phil-abb: Yeah, I see your original concern, and these rules probably do need to be defined so devices don't run into the problem you've identified. 

I don't know how prescriptive the rule needs to be. 

It might be possible to have a requirement that indicates when a device is installing/updating multiple applications as part of the same desired state, they need to ensure their chosen deployment approach does not fail the atomicity check because they didn't take into account other applications being installed as part of the same desired state. We may not need to define how they protect against this, just indicate they need to.

proposals/profile-discovery-and-resolution/01-capability-framework.md
phil-abb: >Given how fast local device activity changes and how often they are reported to WFM, the pre-check wouldn't be accurate.

Why do you feel this will be changing quickly?

I see application deployments occurring very infrequently, except for maybe the first time a device is provisioned and set up with all the desired applications.  Customers are not going to be deploying new apps or updating deployed apps very often. This is not something that is going to be happening multiple times a day or week.

proposals/providers-based-resource-resolution/part3-application-deployment.md
phil-abb: I'm resolving this as an old comment

proposals/providers-based-resource-resolution/part3-application-deployment.md
phil-abb: I'm resovling this as an old comment

proposals/providers-based-resource-resolution/part3-application-deployment.md
phil-abb: I'm resolving this as an old comment

proposals/providers-based-resource-resolution/part1-providers.md
phil-abb: I'm resolving this as an old comment

proposals/providers-based-resource-resolution/part3-application-deployment.md
phil-abb: I'm resolving this as an old comment

proposals/providers-based-resource-resolution/part3-application-deployment.md
phil-abb: I'm resolving this as an old comment

proposals/profile-discovery-and-resolution/01-capability-framework.md
phil-abb: See [this detailed comment](https://github.com/margo/specification-enhancements/pull/66#discussion_r3645778997) about why I'm asking about this.

proposals/profile-discovery-and-resolution/01-capability-framework.md
phil-abb: See [this detailed comment](https://github.com/margo/specification-enhancements/pull/66#discussion_r3645778997) about why I'm asking about this.

proposals/profile-discovery-and-resolution/01-capability-framework.md
phil-abb: @singhmj-1 what do you think about the idea of adding a "validatedBy" property to parameters? This would allow for a profile definition to be either a guard, generator, or both

Here are a couple of examples of what I mean. 

### Network Ports
Instead of having `deviceprofile.margo.org/interfaces/port` guard and `deviceprofile.margo.org/interfaces/dynamic-port` generator, there is just  `deviceprofile.margo.org/interfaces/port`. An application vendor has two options: they can either indicate they want the port to be automatically assigned, or they can say they want the customer to define the port, but they want the value validated before trying to use it.

#### ProfileDefintion:

Based on your dynamic port example

```jason
{
  "profileDefinitionId": "deviceprofile.margo.org/interfaces/port",
  "scope": "device",
  "category": "resource",
  "platformStateSchema": {...},
  "discoverySchema": {
    "mode": ["generator", "guard"]
    "requestSchema": {
      "type": "object",
      "properties": {
        "protocol": { "type": "string" },
        "rangeMin": { "type": "number" },
        "rangeMax": { "type": "number" }
      }
    },
    "outputSchema": {
      "type": "object",
      "properties": {
        "availablePort": { "type": "number" }
      }
    },
    "failureCodes": ["NoPortAvailable"]
  }
}
```

App vendors can define parameters in two ways:

#### ApplicationDescription Parameter - device-assigned port

The application vendor uses the [proposed](https://github.com/margo/specification-enhancements/blob/device-specific-parameter-values/proposals/device-specific-parameter-values.md) "valueFrom" object to indicate they want the parameter value to be provided by the device's `deviceprofile.margo.org/interfaces/port` availablePort output property (generator)

```yaml
parameters:
  servicePort:
    valueFrom:
      device:
        key: deviceprofile.margo.org/interfaces/port/output/availablePort
        required: true
    targets:
      - pointer: global.service.port
        components:
          - fi-web-helm
      - pointer: SERVICE_PORT
        components:
          - fi-web-compose
```

#### ApplicationDescription Parameter - device port validation only

We introduce a new "validatedBy" object to the ApplicationDescription parameters, allowing the application vendors to indicate they want a property to be validated by a device profile (e.g., the `deviceprofile.margo.org/interfaces/port` profile (guard)).

`validatedBy` (or some other name) is optional and allows an app vendor to say how they want the device to validate a parameter if the parameter value needs to be validated by the device (e.g., network port assignment, ingress routes, etc.)

```yaml
parameters:
  servicePort:
    value: 9095
    validatedBy:
      device:
        key: deviceprofile.margo.org/interfaces/port
    targets:
      - pointer: global.service.port
        components:
          - fi-web-helm
      - pointer: SERVICE_PORT
        components:
          - fi-web-compose
configuration:
  sections:
    - name: Network Configuration
      settings:
       - parameter: servicePort
         name: Service Port
         description: Network port to expose the service on
         schema: rangedNumber
   schema: ...
```

### Ingress Route validation

This would allow for the ingress validation case that was part of the original SUP (and original issue)

#### ProfileDefintion

Just a basic example to demonstrate

```json
{
  "profileDefinitionId": "deviceprofile.margo.org/capability/ingress",
  "description": "Profile Definition for Ingress Route configuration",
  "scope": "device",
  "category": "capability",
  "schemaVersion": "1.0.0",
  "platformStateSchema": {...},
   "discoverySchema": {
     "mode": ["guard"]
   },
  "failureCodes": ["RouteUnavailable"]
}
```

#### ApplicationDescription - ingress route validation only

```yaml
parameters:
  ingressPrefix:
    value: northstar
    validatedBy:
      device:
        key: deviceprofile.margo.org/capability/ingress
    targets:
      - pointer: global.ingress.prefix
        components:
          - fi-web-helm
configuration:
  sections:
    - name: Network Configuration
      settings:
       - parameter: ingressPrefix
         name: Ingress Prefix
         description: The prefix to prepend to the hostname to enable access to the application 
         schema: requiredText
   schema: ...
```

jjaswanson4:
Worth discussing: in the resource discovery part of this, we should include the ability to ask k8s devices (technically clusters, but using the Margo term) what capabilities/objects they provide - this can be used to steer deployment of workloads, and/or allow for conditionalizing of various application components when the target deployment is k8s.

For example: (almost) all k[0,3,8,etc]s clusters should provide the "service" object - however ingress to the cluster can differ depending on vendor/deployed components/etc.

Generally, clusters can be asked for all resource types:
```
kubectl api-resources --verbs=create
```

That should generate a list of what objects can be created, CRD or core. I'd argue this should be collected and reported to a WFM to allow for proper application steering - preventing deployments if the application expects a CRD, but the cluster doesn't provide it.

If I have an application deployed by helm chart that wants a route:
```
apiVersion: route.openshift.io/v1
kind: Route
metadata:
...
```
However my target deployment device (cluster) doesn't provide that CRD, the deployment will fail.

This could be prevented by having the WFM match application required objects to target provided - if all the required CRDs/etc exist, then the deployment has a higher chance of succeeding.

In addition, this could be something used for templating of application deployments - a WFM could expose the CRDs/etc for a target device, then allow for conditional rendering of the application deployment.

Using the example above, if my application wants a route by default, but the target deployment device doesn't provide it, then it could fall back to using a service with NodePort.

chrisgclayton:
In reviewing this it has a lot of good elements to it, but I am wondering a bit about the the challenge with distributed aspects. If the plan is to check at the time of deployment, you call out the transient, implicit, exclusive elements. I am looking at this thinking that unless you are "reserving with the first request" you may still fail at deployment depending on WFM behaviour?

PR #67, opened 2026-06-01
docs: SUP-00 The Helm Way — canonical OCI component publishing pattern

javatask:
### Description

  Introduces an informational SUP that names and documents the five-element OCI
  publishing pattern already implemented by Helm in the Margo specification:
  **Storage**, **Reference**, **Media Types**, **Integrity**, and **Parametrization**.

  build upon instead of re-specifying the same five requirements independently. Any
  future component type SUP can demonstrate conformance by filling in the Pattern
  Compliance Table in this document.

  No normative specification changes. No schema modifications. Informational only.
  
  #### Issues Addressed
  
  - Completed SUP: [sup_app_registry_as_oci](https://github.com/margo/specification-enhancements/blob/m  ain/completed/sup_app_registry_as_oci.md) — established the OCI registry model this pattern names
  - [margo/specification #89](https://github.com/margo/specification/issues/89) — OCI registry strategy  (closed)
  - [margo/specification #91](https://github.com/margo/specification/issues/91) — standard approach for  workload manifests (closed)
  
  #### Change Type
  
  - [ ] Fix (change that resolves an issue)
  - [x] New enhancement (change that adds specification content)
  - [ ] Content edits (change that edits existing content)

  #### Checklist
  
  - [x] I have read the [CONTRIBUTING](../CONTRIBUTING.md) document.
  - [x] My changes adhere to the established patterns and best practices.

PR #68, opened 2026-06-01
Compose OCI registry publishing and archive structure

javatask:
### Description

  Applies The Helm Way pattern to Compose components. Closes five
  interoperability gaps in the current single-sentence Compose packaging description:

  1. Mandates OCI registry storage for Compose Archives (`oras push`)
  2. Defines two Margo-specific OCI media types (`compose.v1+json`, `compose.v1.tar.gzip`)
  3. Specifies normative Compose Archive structure (single top-level directory,
     `compose.yaml` required, security constraints against path traversal/symlink escape)
  4. Introduces versioned `compose.v1` deployment profile type (removes bare `compose`)
  
  Since this targets the pre-draft specification, `packageLocation` is removed entirely
  (it has never appeared in a released specification — no migration path needed).

  **Recommended merge order**: Apply The Helm Way first for reference coherence, then this SUP,  then Quadlet SUP, which depends on this SUP's media type table.
  
  #### Change Type
  
  - [ ] Fix (change that resolves an issue)
  - [x] New enhancement (change that adds specification content)
  - [x] Content edits (change that edits existing content)

### DIFF
--- a/proposals/compose-oci/sup-01-compose-oci.md
+++ b/proposals/compose-oci/sup-01-compose-oci.md
@@ -250,7 +250,7 @@ The existing `type` slot already uses `pattern: ^(helm|compose)$`. No change is 
 
 The full normative Compose Archive Structure is defined in Change 2 above (placed in `application-registry.md`). The `application-package.md` file retains only a cross-reference callout pointing to that section.
 
-### Publishing Workflow
+### Publishing Workflow (INFORMATIVE)
 
 To publish a Compose Archive to an OCI-compliant Component Registry, use `oras push` (CNCF project). This is the RECOMMENDED publishing tool for Margo Compose Archives.
 
@@ -313,7 +313,6 @@ If health checks are defined in `compose.yaml`, implementations SHOULD additiona
 | MUST | If `wait` is `true`, the device MUST wait until all containers reach running state. |
 | MUST | If any container exits non-zero during startup, the deployment MUST be reported as failed. |
 | SHOULD | If health checks are defined, implementations SHOULD wait for healthy state. |
-| MUST NOT | `docker compose publish` MUST NOT be used to publish Margo Compose components. |
 | RECOMMENDED | `oras push` is the RECOMMENDED publishing tool for Margo Compose Archives. |
 
 ---
--- a/proposals/compose-oci/sup-01-compose-oci.md
+++ b/proposals/compose-oci/sup-01-compose-oci.md
@@ -1,0 +1,382 @@
+# Compose OCI Registry Publishing and Archive Structure
+
+## Owner
+
+[@javatask](https://github.com/javatask) — Andrii Melashchenko, Belden Inc.
+
+## Summary
+
+This SUP applies The Helm Way pattern (SUP-00) to Compose components. It mandates OCI registry storage for Compose Archives, defines two Margo-specific OCI media types, specifies the normative internal archive structure (single top-level directory, `compose.yaml` required, security constraints), introduces the versioned `compose.v1` deployment profile type, deprecates the `packageLocation` field with a two-phase migration path, and resolves three open "Investigation Needed" blocks in the current specification.
+
+These changes are inseparable: what you push to the registry is defined by the archive structure, so the OCI mandate and the structure specification belong in a single SUP. The result is a complete, interoperable Compose packaging contract — something the current single-sentence description does not provide.
+
+## Reason for proposal
+
+The current specification describes the Compose Archive as "a tarball file containing the `compose.yaml` file and any additional artifacts referenced by the Compose file." This single-sentence description is insufficient for interoperable implementations. Five specific gaps exist:
+
+1. **No OCI registry mandate** — Compose Archives can be served from any URL; there is no integrity guarantee.
+2. **No integrity mechanism** — The spec recommends PGP signing but provides no checkable contract.
+3. **`artifactType` inconsistency** — The Mermaid diagram in `application-registry.md` uses a different string than what implementations use.
+4. **Bare type discriminator** — `"compose"` is not versioned, making future evolution impossible without a flag day.
+5. **Undefined archive structure** — No normative rules exist for directory layout, required files, or security constraints (path traversal, symlink escapes).
+
+Without resolving these gaps, two implementations of the Margo Compose packaging model can produce non-interoperable results while both claiming compliance.
+
+## Requirements alignment acknowledgement
+
+This SUP addresses the following open and recently-closed specification issues:
+
+- **[margo/specification #168](https://github.com/margo/specification/issues/168)** — "Define improvements to compose manifest workloads when targeting typical targets (Docker/Podman)" (open). This SUP defines the normative archive structure and OCI publishing contract that underpins any compose manifest improvement.
+- **[margo/specification #166](https://github.com/margo/specification/issues/166)** — "Define the standard mechanism for compose enabled devices to authenticate and pull Compose Archive" (open). This SUP mandates OCI registry storage, which brings Compose Archives under the same OCI authentication model as application packages (OAuth 2.0 / Bearer token per the existing `application-registry.md` Authentication section).
+- **[margo/specification #179](https://github.com/margo/specification/issues/179)** — "Fix/packagefile location for compose applications" (closed). This SUP replaces `packageLocation` with `repository` + `revision` OCI coordinates and provides the migration path.
+
+**Out of scope (explicitly deferred):**
+- Artifact signing and supply-chain attestation — deferred to a dedicated Margo security SUP.
+- Application dependencies and required infrastructure services (storage, message queues, reverse proxy) — deferred; requires coordination with the Margo Device Interface Working Group.
+
+**Breaking changes are managed** via a two-phase transition. Phase 1 (this SUP): `repository`/`revision` are SHOULD; `packageLocation` remains valid with deprecation warning. Phase 2 (v1-beta1): `repository`/`revision` become REQUIRED; `packageLocation` is removed.
+
+## Technical proposal
+
+### Affected files
+
+| File | Change type |
+|---|---|
+| `system-design/concepts/applications/application-package.md` | Normative update + new section |
+| `system-design/specification/applications/application-registry.md` | Table addition |
+| `src/specification/applications/application-description.linkml.yaml` | Schema update |
+
+---
+
+### Change 1: `application-package.md` — OCI mandate and Investigation Needed resolutions
+
+#### 1a: Replace Compose bullet with OCI-mandated language
+
+**Before** (lines 46–54):
+
+```markdown
+- To target devices, which deploy workloads using Kubernetes, components need to be defined as Helm charts using [Helm (version 3)](https://helm.sh/docs/topics/charts/).
+- To target devices, which deploy workloads using [Compose](https://www.compose-spec.io/), components need to be packaged as [Compose Archives](../../personas-and-definitions/technical-lexicon.md#compose-archive), i.e., a tarball file containing the `compose.yaml` file and any additional artifacts referenced by the Compose file (e.g., configuration files, environment variable files, etc.). Margo recommends to digitally sign this package and to specify the location of the public key in the `ApplicationDescription` (see `keyLocation` [here](../../specification/applications/application-description.md#componentproperties-attributes)). When digitally signing the package PGP encryption MUST be used.
+
+If either one cannot be implemented it MAY be omitted but Margo RECOMMENDS defining [deployment profiles](../../specification/applications/application-description.md#deploymentprofile-attributes) as both Helm chart **AND** Compose components to strengthen interoperability and applicability.
+
+> **Note**
+> A device running the application will only install the application using either the Compose Archives or the Helm Charts, but not both.
+```
+
+**After**:
+
+```markdown
+- To target devices which deploy workloads using Kubernetes, components MUST be defined as Helm charts using [Helm (version 3)](https://helm.sh/docs/topics/charts/). Helm charts MUST be stored in an OCI-compliant [Component Registry](../../personas-and-definitions/technical-lexicon.md#component-registry) and referenced via `repository` (an `oci://` URI) and `revision` (an OCI tag) in the [ApplicationDescription](../../specification/applications/application-description.md#componentproperties-attributes).
+- To target devices which deploy workloads using [Compose](https://www.compose-spec.io/), the following requirements apply:
+  1. Components MUST be packaged as [Compose Archives](../../personas-and-definitions/technical-lexicon.md#compose-archive) (see [Compose Archive Structure](#compose-archive-structure) below).
+  2. Compose Archives MUST be stored in an OCI-compliant [Component Registry](../../personas-and-definitions/technical-lexicon.md#component-registry).
+  3. The Compose Archive MUST be pushed to the registry as an OCI artifact (e.g., using `oras push` or equivalent tooling) and referenced via `repository` (an `oci://` URI) and `revision` (an OCI tag) in the [ApplicationDescription](../../specification/applications/application-description.md#componentproperties-attributes).
+  4. The OCI image manifest for the artifact MUST use `application/vnd.org.margo.component.compose.v1+json` as its `artifactType`.
+  5. The layer blob containing the tarball MUST use `application/vnd.org.margo.component.compose.v1.tar.gzip` as its `mediaType`.
+  6. Integrity of the Compose Archive is provided by the OCI content-addressable digest as mandated by the [OCI Distribution Specification v1.1.0](https://github.com/opencontainers/distribution-spec/blob/v1.1.0/spec.md), which the Margo Application Registry already requires.
+
+If any one component type cannot be implemented it MAY be omitted, but Margo RECOMMENDS defining [deployment profiles](../../specification/applications/application-description.md#deploymentprofile-attributes) using multiple component types (Helm, Compose, and/or Quadlet) to strengthen interoperability and applicability across heterogeneous edge device fleets.
+
+> **Note**
+> A device running the application will only install the application using one component type (Helm Charts, Compose Archives, or Quadlet Archives), never more than one simultaneously.
+```
+
+#### 1b: Resolve three "Investigation Needed" blocks
+
+**Before**:
+
+```markdown
+Margo will provide more detailed discussion and specification on the following points:
+
+> **Investigation Needed**: Question: do we need to specify the location of a SHA256 hash for the Compose Archive also (similar to the PGP key) in the ApplicationDescription?
+> We will also discuss how we should handle secure container registries that require a username and password.
+>
+> **Investigation Needed**: We need to determine what impact, if any, using 3rd party helm charts has on being Margo compliant.
+>
+> **Investigation Needed**: Missing in the current specification are ways to define dependencies (e.g., application dependencies) as well as required infrastructure services such as storage, message queues/bus, reverse proxy, or authentication/authorization/accounting.
+```
+
+**After**:
+
+```markdown
+> **Resolved (SUP-01)**: The integrity question is resolved by mandating OCI registry storage for all component types. Integrity verification is provided by the OCI content-addressable digest as mandated by the [OCI Distribution Specification v1.1.0](https://github.com/opencontainers/distribution-spec/blob/v1.1.0/spec.md), which the Margo Application Registry already requires. No separate `digest` field is needed in the ApplicationDescription. Authentication for secure container registries SHOULD follow the mechanisms defined in the [Authentication, Authorization & Security](../../specification/applications/application-registry.md#authentication-authorization--security) section.
+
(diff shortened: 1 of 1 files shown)


phil-abb:
@margo/technical-wg A new SUP has been started by @javatask to define the Compose OCI registry publishing and archive structure. 

This SUP is ready for final review. Please review the proposal and add any questions/comments to the PR.

proposals/compose-oci/sup-01-compose-oci.md
```diff
@@ -0,0 +1,341 @@
+# Compose OCI Registry Publishing and Archive Structure
+
+## Owner
+
+[@javatask](https://github.com/javatask) — Andrii Melashchenko, Belden Inc.
+
+## Summary
+
+This SUP applies The Helm Way pattern (SUP-00) to Compose components. It mandates OCI registry storage for Compose Archives, defines two Margo-specific OCI media types, specifies the normative internal archive structure (single top-level directory, `compose.yaml` required, security constraints), introduces the versioned `compose.v1` deployment profile type, removes the `packageLocation` field (which has never appeared in a released specification), and resolves three open "Investigation Needed" blocks in the current specification.
```
phil-abb: Is there a specific reason you want to include `.v1` with the type `compose.v1`? 

I mainly wanted to make sure you were not just adding it because we previously had `helm.v3`, which was recently changed to just `helm` with the acceptance of [this SUP](https://github.com/margo/specification-enhancements/blob/main/completed/sup_add_helm_v4_support.md#specification-updates)

proposals/compose-oci/sup-01-compose-oci.md
```diff
@@ -0,0 +1,341 @@
...
+>
+> **Investigation Needed**: We need to determine what impact, if any, using 3rd party helm charts has on being Margo compliant.
+>
+> **Investigation Needed**: Missing in the current specification are ways to define dependencies (e.g., application dependencies) as well as required infrastructure services such as storage, message queues/bus, reverse proxy, or authentication/authorization/accounting.
+```
+
+**After**:
+
+```markdown
+> **Resolved (SUP-01)**: The integrity question is resolved by mandating OCI registry storage for all component types. Integrity verification is provided by the OCI content-addressable digest as mandated by the [OCI Distribution Specification v1.1.0](https://github.com/opencontainers/distribution-spec/blob/v1.1.0/spec.md), which the Margo Application Registry already requires. No separate `digest` field is needed in the ApplicationDescription. Authentication for secure container registries SHOULD follow the mechanisms defined in the [Authentication, Authorization & Security](../../specification/applications/application-registry.md#authentication-authorization--security) section.
```
phil-abb: I'm not sure if this literally means updating the note in the specification or not, but if it does, I don't think we need to do that. We can just remove the note.

proposals/compose-oci/sup-01-compose-oci.md
```diff
@@ -0,0 +1,341 @@
...
+
+> **Resolved (SUP-01)**: Using 3rd-party Helm charts as components in an Application Package does not affect Margo compliance of the Application Package itself. Margo compliance is determined by the conformance of the ApplicationDescription and its packaging within the Application Registry. The 3rd-party chart need not be "Margo-aware"; however, the application developer is responsible for ensuring the chart functions correctly on Margo-compliant edge devices. The `pointer` field in Parameter Targets works with any Helm chart's `values.yaml` structure. Application developers SHOULD document known limitations of 3rd-party charts (e.g., charts requiring CRDs, admission webhooks, or cluster-scoped resources) in the deployment profile `description` field.
```
phil-abb: I'm not sure if this literally means updating the note in the specification or not, but if it does, I don't think we need to do that. We can just remove the note.

proposals/compose-oci/sup-01-compose-oci.md
```diff
@@ -0,0 +1,341 @@
...
+
+> **Deferred**: Defining application dependencies and required infrastructure services (storage, message queues, reverse proxy, authentication/authorization/accounting) is deferred to a separate SUP. This topic requires coordination with the Margo Device Interface working group.
```
phil-abb: I'm not sure if we really need to keep this note either. We have some of this in place with the "requiredResources" property in the application description and the device capabilities file from the device. It doesn't cover everything, but I think the TWG is aware of this.

proposals/compose-oci/sup-01-compose-oci.md
```diff
@@ -0,0 +1,341 @@
...
+      - `quadlet.v1`: Quadlet-based component (see QuadletDeploymentProfile, introduced by SUP-02).
+    rank: 10
+    range: string
+    required: true
+    pattern: ^(helm\.v3|compose\.v1|quadlet\.v1)$
+```
+
+---
+
+### Change 4: New section `## Compose Archive Structure` in `application-package.md`
```
phil-abb: The "application-package" page is not part of the specification content, so it shouldn't contain spec language (MUST, MAY, SHALL). Also, the application package page is talking about the package that contains the application description content, not the package for the deployed artifacts.

 I think this content either needs to go on a new page under the `Specification/Applications` section or into the [application registry](https://docs.margo.org/specification/applications/application-registry) page which has similar content for Helm.

The reason for this is we don't want to have to version the non-specification content (everything that's not under the Specification folder), so we want to keep the spec content seperate from the general content. 

I think the current "application package" page content is violating this principle because it currently has spec language.

Silvanoc:
Since Docker Compose has since version 2.34.0 native support for it: https://docs.docker.com/compose/how-tos/oci-artifact/

I don't like the Docker Compose design for OCI packaging that much (see [1](https://github.com/docker/compose/pull/13257#issuecomment-3393642833) and [2](https://github.com/docker/compose/issues/13354)), but it probably makes no sense to build our own alternative...

proposals/compose-oci/sup-01-compose-oci.md
matlec: I got the same question / doubt while reviewing the SUP. The compose specification actually dropped the `version` property and deliberately moved away from versioning (see [here](https://github.com/compose-spec/compose-spec/blob/main/04-version-and-name.md)).

proposals/compose-oci/sup-01-compose-oci.md
```diff
@@ -0,0 +1,341 @@
...
+#### Margo-Specific Media Types
+
+|Media Type|Description|
+|----------|----------|
+|``application/vnd.org.margo.app.v1+json`` | MUST be used as the **artifactType** to mark the OCI image manifest as the definition of a Margo Application Package |
+|``application/vnd.org.margo.app.description.v1+yaml`` | MUST be used to mark a layer in the OCI image manifest as pointing to the Margo Application Description file |
+|``application/vnd.org.margo.app.icon.v1+{file format}``| MUST be used to mark a layer in the OCI image manifest as pointing to the icon of a Margo Application Package |
+|``application/vnd.org.margo.app.descriptionFile.v1+{file format}``| MUST be used to mark a layer in the OCI image manifest as pointing to description file of a Margo Application Package |
+|``application/vnd.org.margo.app.licenseFile.v1+{file format}``| MUST be used to mark a layer in the OCI image manifest as pointing to the license file of a Margo Application Package|
+|``application/vnd.org.margo.app.releaseNotes.v1+{file format}``| MUST be used to mark a layer in the OCI image manifest as pointing to the release notes file of a Margo Application Package|
```
matlec: See my comment in the [SUP-00](https://github.com/margo/specification-enhancements/pull/67#discussion_r3347531871), this diff is not a true diff against the current spec.

proposals/compose-oci/sup-01-compose-oci.md
```diff
@@ -0,0 +1,341 @@
...
+
+##### Component Registry Media Types
+
+|Media Type|Description|
+|----------|----------|
+|``application/vnd.org.margo.component.compose.v1+json``| MUST be used as the **artifactType** to mark an OCI image manifest as a Margo Compose Archive component in a Component Registry |
+|``application/vnd.org.margo.component.compose.v1.tar.gzip``| MUST be used as the **mediaType** for the layer blob containing the Compose Archive tarball (.tar.gz) |
```
matlec: See my comment in the [SUP-00](https://github.com/margo/specification-enhancements/pull/67#discussion_r3347531871) regarding `tar+gzip`.

proposals/compose-oci/sup-01-compose-oci.md
```diff
@@ -0,0 +1,341 @@
...
+        range: string
+        required: true
+      revision:
+        description: >-
+          OCI tag identifying the component version within the repository (e.g., "1.0.0", "2.3.1").
+          SemVer 2.0 version without leading `v`.
+          MUST be used for Helm, Compose, and Quadlet components.
+        rank: 20
+        range: string
+        required: true
```
matlec, line 182: I think we should define a Regex for the `revision` property. This is especially needed since regular SemVer 2.0 versions cannot be used as-is as mentioned in my [other comment](https://github.com/margo/specification-enhancements/pull/67#discussion_r3347439272). We should either drop the support for build metadata or apply the work-around Helm chose.

proposals/compose-oci/sup-01-compose-oci.md
```diff
@@ -0,0 +1,341 @@
...
+      wait:
+        description: If True, indicates the device waits for the component installation to complete.
+        rank: 30
+        range: boolean
```
matlec, line 188: We might want to be more specific here how `wait` maps to implementations (e.g. Docker Compose). Would `wait: true` mean running synchronous `docker compose up` with a process timeout and waiting for containers to reach the `running` state, or `docker compose up --wait` which waits for containers reaching a `healthy` state (given that health checks have been defined)? For Helm this ambiguity does not exist because `--wait` is precisely defined by the Helm CLI. Compose has no single spec-level equivalent.

proposals/compose-oci/sup-01-compose-oci.md
```diff
@@ -0,0 +1,341 @@
...
+
+### Directory Layout
+
+The archive MUST contain exactly one top-level directory. The name of this directory MUST match the component `name` as specified in the ApplicationDescription.
+
+The top-level directory MUST contain a file named `compose.yaml`. The Compose file MUST conform to the [Compose Specification](https://www.compose-spec.io/) as currently published.
+
+> **Note:** The Compose file MUST be named `compose.yaml`. The alternative names `compose.yml`, `docker-compose.yaml`, and `docker-compose.yml` are NOT valid within a Margo Compose Archive. This restriction ensures predictable file discovery by WFM and device implementations.
+
+All files referenced by `compose.yaml` (including but not limited to `env_file` entries, `volumes` bind-mount sources, `configs` file sources, and `secrets` file sources) MUST be included within the archive and MUST be referenced using relative paths that resolve within the top-level directory.
```
matlec: A compose file might define bind mounts for runtime data, e.g.,

```
volumes:
  - ./data:/var/lib/app/data
```

For those bind mounts, I wouldn't expect `data` to be part of the archive.

matlec: I disagree here: files used to resolve `secrets` should certainly NOT be part of the tarball. Pushing secrets to an OCI registry is not a good idea for many reasons. We need a different mechanism for resolving secrets in Compose archives.

proposals/compose-oci/sup-01-compose-oci.md
```diff
@@ -0,0 +1,341 @@
...
+Add after the existing component type list:
+
+````markdown
+## Compose Archive Structure
+
+A Compose Archive is a gzip-compressed tar archive (`.tar.gz` or `.tgz`) that packages a [Compose](https://www.compose-spec.io/) application for deployment on edge devices. The archive MUST conform to the following structural requirements.
+
+### Directory Layout
+
+The archive MUST contain exactly one top-level directory. The name of this directory MUST match the component `name` as specified in the ApplicationDescription.
```
matlec: > The name of this directory MUST match the component `name` as specified in the ApplicationDescription.

Why would we need to define such a constraint? If the archive MUST contain exactly one top-level directory, the name of that directory doesn't even matter. Implementations can always find the `compose.yaml` by entering whatever that single directory is. This requirement (imho) adds unnecessary coupling between a Compose component and a Margo app package. When the author of an app package decides to rename the component, the component would need to be re-published.

proposals/compose-oci/sup-01-compose-oci.md
javatask: We agree. The .v1 suffix was added by analogy with the old helm.v3 convention, which has since been removed.

proposals/compose-oci/sup-01-compose-oci.md
javatask: Agreed.

proposals/compose-oci/sup-01-compose-oci.md
javatask: Agreed.

proposals/compose-oci/sup-01-compose-oci.md
javatask: Agreed.

proposals/compose-oci/sup-01-compose-oci.md
javatask: Agreed that normative language must not live in concepts/. We will move the Compose Archive Structure MUST-level text to application-registry.md.

However, we ask that application-package.md retain an explicit normative cross-reference callout — not just a prose link — so that implementors reading the packaging documentation are unambiguously directed to the binding requirements. The payload format is half the interoperability contract; it must not appear optional to someone reading the concepts section.

Concretely, application-package.md will include the following callout after the Compose Archive paragraph:

Normative requirement: The internal structure of a Compose Archive is normatively defined in Compose Archive Structure. Implementations MUST conform to that definition.

proposals/compose-oci/sup-01-compose-oci.md
javatask: Agreed

proposals/compose-oci/sup-01-compose-oci.md
javatask: Agreed. The revision field MUST follow the Helm convention: SemVer + (build metadata separator) MUST be replaced with _ since + is not a valid OCI tag character.

Proposed pattern constraint for ComponentProperties.revision in application-description.linkml.yaml and desired-state.linkml.yaml:

```
 description: >-
    OCI tag identifying the component version within the repository
    (e.g., "1.0.0", "2.3.1", "1.0.0-rc1", "1.0.0_build.123").
    MUST be a valid SemVer 2.0 string without a leading `v`.
    SemVer build metadata separator `+` MUST be stored as `_`
    because `+` is not a valid OCI tag character.
    Implementations comparing this value against a SemVer string
    MUST convert `_` back to `+` before comparison.
  range: string
  pattern: "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:_([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$"
```

proposals/compose-oci/sup-01-compose-oci.md
javatask: wait: true MUST mean all containers in the Compose project reach running state — equivalent to synchronous docker compose up or podman-compose up completing without --detach. If any container exits with a non-zero code during startup, the deployment MUST report failure immediately.

Health-check-based readiness is a SHOULD: implementations SHOULD wait for health checks to pass if defined in compose.yaml, but this cannot be MUST because not all Compose files define health checks and mandating it would force all Compose archives to add them.

proposals/compose-oci/sup-01-compose-oci.md
javatask: Correct. Revised rule:
Files referenced by compose.yaml via env_file, configs (file source), and secrets (file source) MUST be included in the archive. Bind-mount volume paths declared in volumes are runtime paths and MUST NOT be included in the archive.

javatask: Agreed. Revised:

secrets with file: source → MUST NOT be in the archive
The Compose file SHOULD use environment: secrets (injected at runtime by the device/WFM) rather than file-based secrets
Secret provisioning is out of scope for this SUP

proposals/compose-oci/sup-01-compose-oci.md
javatask: The single top-level directory requirement follows the Helm precedent: a Helm chart directory name must match the chart name, and helm package enforces this. We applied the same pattern to Compose for consistency across all Margo component types.

What stays MUST — the structural constraint: The archive MUST contain exactly one top-level directory. This is the interoperability-critical rule. It ensures a deterministic extraction path for any WFM or device implementation regardless of the tool used to unpack the archive. It also mirrors Helm's structural requirement.

What relaxes to SHOULD — the naming constraint: The directory name SHOULD match the component name for human readability and debugging, but implementations MUST NOT depend on the directory name for discovery. Discovery algorithm: enter the single top-level directory; find compose.yaml.

On subdirectories: The "exactly one top-level directory" rule does not restrict the contents of that directory. Applications requiring multiple folders — configs/, certs/, scripts/ alongside compose.yaml — are fully supported. All such subdirectories live inside the single top-level directory. This is the same model Helm uses for templates/, charts/, crds/ inside the chart root.

javatask:
@Silvanoc 
he picture is split across two layers:
 
| Layer | Repository | License | Roadmap control |
|-------|-----------|---------|----------------|
| **Compose Specification** (the format) | `compose-spec/compose-spec` | Apache 2.0 | `compose-spec` GitHub org — open governance with Docker, AWS, Microsoft and others as stated contributors |
| **Docker Compose** (the CLI implementation) | `docker/compose` | Apache 2.0 | `docker/` GitHub org — Docker Inc. controls the roadmap; `docker compose publish` was added and designed here |
| **ORAS** (OCI push/pull tool) | `oras-project/oras` | Apache 2.0 | CNCF project under vendor-neutral governance |
 
The *file format* (compose-spec) has genuine open governance. The *`docker compose publish` feature* is in `docker/compose`, which Docker Inc. controls. Docker explicitly stated in 2020 its intent to submit the spec to an open source foundation — that has not happened. The reference implementation remains `docker/compose` under Docker Inc.'s roadmap. 
https://www.docker.com/blog/announcing-the-compose-specification/
 
**The real question**: should Margo reference `docker compose publish` as a normative publishing tool?
 
**Options and tradeoffs**:
 
**Option A — Reference `docker compose publish` as a normative or co-equal tool**
 
- Pro: converges with what developers already use; `docker compose publish` produces a valid OCI artifact; reduces friction for Docker-heavy environments
- Con: `docker compose publish` is implemented in `docker/compose` (Docker Inc. roadmap); its OCI design has known concerns (Silvanoc's own links); it is not available in Podman, nerdctl, or other Compose-compatible runtimes — mandating it would exclude any non-Docker publishing pipeline
- Con: the spec would have an asymmetry: the *file format* is governed by compose-spec, but the *publishing mechanism* would be governed by Docker Inc.

**Option B — Reference `oras push` as the normative tool; explicitly allow `docker compose publish` as a conforming alternative**
 
- Pro: `oras push` is a CNCF project, vendor-neutral, works with Docker and Podman alike, and is already the reference tool for OCI artifact publishing in the CNCF ecosystem
- Pro: Margo's logistics layer matches its governance model (LF/CNCF tooling for an LF specification)
- Pro: `docker compose publish` is not excluded — it MAY be used if the output conforms to Margo's media types and archive structure
- Con: slightly more friction for Docker Desktop users who expect `docker compose publish` to "just work" with Margo

**However — `docker compose publish` produces an incompatible artifact structure**: Investigation of the `docker/compose` source shows that `docker compose publish` sets `artifactType: application/vnd.docker.compose.project` and produces **multiple separate OCI layers** — one per Compose file, one per env file — with file paths replaced by SHA256 hashes. This is structurally incompatible with the Margo archive format, which requires a **single layer blob** containing a `tar+gzip` archive with `mediaType: application/vnd.org.margo.component.compose.tar+gzip`. A WFM or device implementation following the Margo spec cannot consume a `docker compose publish` artifact without custom parsing logic.
 
Therefore `docker compose publish` is **not** a conforming alternative for Margo Compose Archive publishing. `oras push` with a pre-built `tar.gz` is the only reference tool.

javatask:
@phil-abb , @matlec , @Silvanoc, the new version of SUP with your improvements was published

javatask:
Hello @pperycz , @lucamaf, @matlec, @MariusKnobbe, @cplatzer-se , @PankajShaw-SE , @zwigby , @acdaniel , @achuza , @OMuhammad , @matlec , @MariusKnobbe plese use this SUP as the base for compose OCI testing.
[README-margo-community.md](https://github.com/user-attachments/files/29472795/README-margo-community.md)

phil-abb:
This SUP has been approved, so force-merge SUP: https://github.com/margo/specification-enhancements/issues/80

PR #69, opened 2026-06-01
Quadlet component type (Podman 5.0+ / systemd)

javatask:
### Description

  Adds Quadlet as a third Margo component type alongside Helm and Compose, following The Helm Way pattern. Quadlet (Podman 5.0+) provides native systemd
  integration for container workloads without a Kubernetes control plane — the
  lightweight, daemon-free deployment model required on constrained OT edge devices.

  This SUP is **additive only** — no existing Helm or Compose functionality is changed.

  Changes:
  - `quadlet.v1` deployment profile type + `QuadletDeploymentProfile` / `QuadletComponent` schema
  classes
  - Two new Margo-registered OCI media types (`quadlet.v1+json`, `quadlet.v1.tar.gzip`)
  - Normative Quadlet Archive Structure (single-container and multi-container examples,
  - Companion patch: `PATCH_desired-state-quadlet.diff` — adds Quadlet classes to
    `desired-state.linkml.yaml` (required for WFM to generate schema-valid desired-state
    documents for Quadlet workloads)

  > **Dependency**: This SUP MUST be applied after Compose OCI Registry
  > Publishing, which establishes the Component Registry media type table and type
  > slot regex that this SUP extends.

    #### Change Type
  
  - [ ] Fix (change that resolves an issue)
  - [x] New enhancement (change that adds specification content)
  - [] Content edits (change that edits existing content)

### DIFF
--- a/proposals/quadlet/PATCH_desired-state-quadlet.diff
+++ b/proposals/quadlet/PATCH_desired-state-quadlet.diff
@@ -15,8 +15,8 @@
 +  QuadletDeploymentProfile:
 +    is_a: DeploymentProfile
 +    description: >-
-+      Deployment profile for Quadlet-based workloads targeting devices running
-+      Podman 5.0 or later with systemd integration.
++      Deployment profile for Quadlet-based workloads targeting devices capable
++      of processing Quadlet unit files via systemd integration.
 +    rank: 45
 +    slot_usage:
 +      type:
@@ -37,16 +37,16 @@
 +    is_a: Component
 +    rank: 55
 +    description: >-
-+      A component packaged as a Quadlet Archive for deployment on devices using
-+      Podman with systemd integration.
++      A component packaged as a Quadlet Archive for deployment on devices
++      capable of processing Quadlet unit files via systemd integration.
 +
 
    type:
 -    description: The type of deployment profile (e.g., helm, compose).
 +    description: >-
 +      The type of deployment profile. Allowed values are: helm (Kubernetes/Helm
-+      chart deployment), compose (Compose file deployment via Podman or Docker
-+      Compose), and quadlet (Podman systemd Quadlet unit file deployment). The
++      chart deployment), compose (Compose file deployment), and quadlet
++      (Quadlet unit file deployment via systemd integration). The
 +      value MUST match the type field in the corresponding ApplicationDescription
 +      deploymentProfile element.
      range: string
--- a/proposals/quadlet/PATCH_desired-state-quadlet.diff
+++ b/proposals/quadlet/PATCH_desired-state-quadlet.diff
@@ -1,0 +1,108 @@
+# PATCH: desired-state.linkml.yaml -- Add QuadletDeploymentProfile and QuadletComponent
+#
+# Schema file:
+#   src/specification/margo-management-interface/desired-state.linkml.yaml
+#
+# Purpose:
+#   Close R3 gap identified by margo-tech-reviewer in REVIEW_tech-reviewer.md.
+#   Without these additions, WFMs cannot generate schema-valid
+#   ApplicationDeployment desired-state documents for Quadlet workloads.
+#   The type slot accepts the string "quadlet" (no pattern constraint exists
+#   in the desired-state schema) but there is no QuadletDeploymentProfile or
+#   QuadletComponent class to validate against, leaving Quadlet desired-state
+#   documents effectively untyped.
+#
+# How to apply:
+#   This unified diff is NOT committed directly to the submodule. To apply:
+#
+#       cd documentation/spec_analysis/specification
+#       patch -p1 < ../proposals/PATCH_desired-state-quadlet.diff
+#
+#   After applying, run the validation toolchain:
+#
+#       bash doc-generation/check-examples.bash
+#
+# Structural rationale:
+#   In desired-state.linkml.yaml the Component class uses
+#   "range: Property" (a name/value pair list, lines 151-158), NOT
+#   "range: ComponentProperties" as in application-description.linkml.yaml.
+#   This reflects the different role of the two schemas: the desired-state
+#   document carries runtime parameter injection (name/value pairs injected
+#   by the WFM) whereas the ApplicationDescription carries OCI package
+#   registry coordinates (ComponentProperties with repository/revision).
+#
+#   QuadletComponent therefore inherits from the same Component base class
+#   as HelmComponent and ComposeComponent, using the name/value Property
+#   model. No new Property subclass is needed.
+#
+#   The type slot description is updated to document the three accepted
+#   values. No pattern constraint is added here (consistent with the
+#   existing schema which omits a pattern on the desired-state type slot).
+#   Adding a pattern would be a separate, follow-up schema hardening change.
+#
+# Dependency:
+#   This patch is a companion to WG-PROPOSAL-01. It MUST be applied
+#   after (or simultaneously with) the application-description.linkml.yaml
+#   changes from that proposal.
+
+--- a/src/specification/margo-management-interface/desired-state.linkml.yaml
++++ b/src/specification/margo-management-interface/desired-state.linkml.yaml
+@@ -130,6 +130,19 @@ classes:
+   ComposeDeploymentProfile:
+     is_a: DeploymentProfile
+     rank: 40
+     slot_usage:
+       type:
+         equals_string: "compose"
+         rank: 10
+       components:
+         range: ComposeComponent
+         rank: 20
+
++  QuadletDeploymentProfile:
++    is_a: DeploymentProfile
++    description: >-
++      Deployment profile for Quadlet-based workloads targeting devices running
++      Podman 5.0 or later with systemd integration. Components are referenced
++      via repository (an oci:// URI) and revision (an OCI tag) consistent with
++      the Helm and Compose OCI publishing model mandated by WG-PROPOSAL-01.
++    rank: 40
++    slot_usage:
++      type:
++        equals_string: "quadlet"
++        rank: 10
++      components:
++        range: QuadletComponent
++        rank: 20
++
+   Component:
+     description: A class representing a component of a deployment profile.
+     rank: 50
+@@ -160,6 +173,12 @@ classes:
+   ComposeComponent:
+     is_a: Component
+     rank: 50
+
++  QuadletComponent:
++    is_a: Component
++    rank: 50
++    description: >-
++      A component packaged as a Quadlet Archive for deployment on devices using
++      Podman with systemd integration. Properties follow the same name/value
++      pair model as HelmComponent and ComposeComponent in this schema.
++
+   Property:
+     rank: 80
+     attributes:
+@@ -213,7 +232,11 @@ slots:
+   type:
+-    description: The type of deployment profile (e.g., helm.v3, compose).
++    description: >-
++      The type of deployment profile. Allowed values are: helm.v3 (Kubernetes/Helm
++      chart deployment), compose (Compose file deployment via Podman or Docker
++      Compose), and quadlet (Podman systemd Quadlet unit file deployment). The
++      value MUST match the type field in the corresponding ApplicationDescription
++      deploymentProfile element.
+     range: string
+     required: true
+     rank: 10
--- a/proposals/quadlet/PATCH_desired-state-quadlet.diff
+++ b/proposals/quadlet/PATCH_desired-state-quadlet.diff
@@ -1,53 +1,6 @@
-# PATCH: desired-state.linkml.yaml -- Add QuadletDeploymentProfile and QuadletComponent
-#
-# Schema file:
-#   src/specification/margo-management-interface/desired-state.linkml.yaml
-#
-# Purpose:
-#   Close R3 gap identified by margo-tech-reviewer in REVIEW_tech-reviewer.md.
-#   Without these additions, WFMs cannot generate schema-valid
-#   ApplicationDeployment desired-state documents for Quadlet workloads.
-#   The type slot accepts the string "quadlet" (no pattern constraint exists
-#   in the desired-state schema) but there is no QuadletDeploymentProfile or
-#   QuadletComponent class to validate against, leaving Quadlet desired-state
-#   documents effectively untyped.
-#
-# How to apply:
-#   This unified diff is NOT committed directly to the submodule. To apply:
-#
-#       cd documentation/spec_analysis/specification
-#       patch -p1 < ../proposals/PATCH_desired-state-quadlet.diff
-#
-#   After applying, run the validation toolchain:
-#
-#       bash doc-generation/check-examples.bash
-#
-# Structural rationale:
-#   In desired-state.linkml.yaml the Component class uses
-#   "range: Property" (a name/value pair list, lines 151-158), NOT
-#   "range: ComponentProperties" as in application-description.linkml.yaml.
-#   This reflects the different role of the two schemas: the desired-state
-#   document carries runtime parameter injection (name/value pairs injected
-#   by the WFM) whereas the ApplicationDescription carries OCI package
-#   registry coordinates (ComponentProperties with repository/revision).
-#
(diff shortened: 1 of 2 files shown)


phil-abb:
@margo/technical-wg A new SUP has been started by @javatask to define Quadlet deployment type. 

This SUP is ready for final review. Please review the proposal and add any questions/comments to the PR.

proposals/quadlet/sup-02-quadlet.md
```diff
@@ -0,0 +1,325 @@
...
+
+## Owner
+
+[@javatask](https://github.com/javatask) — Andrii Melashchenko, Belden Inc.
+
+## Summary
+
+This SUP adds Quadlet as a third Margo component type alongside Helm and Compose. Quadlet is Podman's native systemd integration format for container workloads — it allows containers to be managed as native systemd services without requiring a Kubernetes control plane. This is the lightweight, daemon-free deployment model required on constrained OT edge devices.
+
+The SUP applies The Helm Way pattern (SUP-00) to Quadlet, following the same OCI registry publishing model established by SUP-01 for Compose. It defines the `quadlet.v1` deployment profile type, two Margo-registered OCI media types, `QuadletDeploymentProfile` and `QuadletComponent` schema classes, the normative Quadlet Archive Structure, and a companion patch for the desired-state schema. This SUP is **additive only** — no existing Helm or Compose functionality is changed.
```
phil-abb: Is there a specific reason you want to include .v1 with the quadlet.v1?

I mainly wanted to make sure you were not just adding it because we previously had helm.v3, which was recently changed to just helm with the acceptance of [this SUP](https://github.com/margo/specification-enhancements/blob/main/completed/sup_add_helm_v4_support.md#specification-updates)

proposals/quadlet/sup-02-quadlet.md
```diff
@@ -0,0 +1,325 @@
...
+
+The current Margo specification has no path for systemd-managed container workloads. Edge devices running Podman 5.0+ with systemd cannot be managed by a Margo-compliant WFM today. This gap is significant: many industrial OT edge devices run a full Linux distribution with systemd but cannot justify the resource overhead of a Kubernetes control plane.
+
+Quadlet is production-ready and already deployed in industrial environments. Adding it as a first-class Margo component type closes the gap between the Margo application model and the reality of OT edge device deployments.
+
+## Requirements alignment acknowledgement
+
+This SUP is grounded in the following in-scope features and issues:
+
+- **[margo/specification-enhancements #60](https://github.com/margo/specification-enhancements/issues/60)** — "Mandate daemonless container runtime as the default single edge container runtime for Margo-compliant edge devices" (open). Quadlet/Podman is the reference daemonless container runtime. This SUP provides the Margo packaging contract for Quadlet workloads on such devices.
```
phil-abb: I'm not sure if we should be referencing this enhancement request. This has come from someone outside of Margo and hasn't been reviewed with the PM group as being within the Scope of Margo. My opinion is that this enhancement request goes beyond Margo's scope, but we'll have to wait for a PM review/decision.

proposals/quadlet/sup-02-quadlet.md
```diff
@@ -0,0 +1,325 @@
...
+- **[margo/specification #168](https://github.com/margo/specification/issues/168)** — "Define improvements to compose manifest workloads when targeting typical targets (Docker/Podman)" (open). Quadlet complements Compose for Podman-based targets: Compose targets Docker-compatible runtimes; Quadlet targets systemd-integrated Podman deployments where lifecycle management (restart, dependency ordering, journald logging) is required.
+
```
phil-abb, line 30: There is currently no feature specifically for supporting Quadlets. I think this is something that needs to be brought up with the PM group and a feature created for it. I don't expect there will be any concerns with supporting this additional deployment type.

[reference](https://github.com/margo/Margo-collaboration/blob/main/specification_update_processes/process-overview.md#phase-3-sup-technical-design):
> During this phase, the SUP owner must ensure a feature exists in the TWG Feature Project and is listed in the "Requirements Alignment Acknowledgement" section. If a feature does not exist, work with the TWG Chair/Co-Chair to ensure a feature is created

proposals/quadlet/sup-02-quadlet.md
```diff
@@ -0,0 +1,325 @@
...
+### Affected files
+
+| File | Change type |
+|---|---|
+| `system-design/concepts/applications/application-package.md` | Normative update + new section |
+| `system-design/specification/applications/application-registry.md` | Table addition (extends SUP-01 table) |
+| `src/specification/applications/application-description.linkml.yaml` | Schema addition |
+| `src/specification/margo-management-interface/desired-state.linkml.yaml` | Schema addition (companion patch) |
+
```
phil-abb, line 50: There might be another update that's needed. If the [SUP to remove roles from the device capabilities](https://github.com/margo/specification-enhancements/pull/50) is approved, it adds the `supportedDeploymentTypes` array, so quadlets would need to be added to that.

proposals/quadlet/sup-02-quadlet.md
```diff
@@ -0,0 +1,325 @@
...
+      artifactType application/vnd.org.margo.component.quadlet.v1+json
+      (see application-registry.md).
+    #rank: 77
+```
+
+#### 3b: Expand `type` slot regex
+
+The `type` slot pattern introduced by SUP-01 is:
+```yaml
+    pattern: ^(helm\.v3|compose|compose\.v1)$
```
phil-abb: As indicated in the note above, "helm.v3" was changed to "helm"

proposals/quadlet/sup-02-quadlet.md
```diff
@@ -0,0 +1,325 @@
...
+Update to:
+```yaml
+    pattern: ^(helm\.v3|compose|compose\.v1|quadlet\.v1)$
+```
+
+> **Note:** SUP-01 Change 3c already includes `quadlet.v1` in the `type` slot description text. Only the regex requires expansion here.
+
+---
+
+### Change 4: New section `## Quadlet Archive Structure` in `application-package.md`
```
phil-abb: The "application-package" page is not part of the specification content, so it shouldn't contain spec language (MUST, MAY, SHALL). Also, the application package page is talking about the package that contains the application description content, not the package for the deployed artifacts.

 I think this content either needs to go on a new page under the `Specification/Applications` section or into the [application registry](https://docs.margo.org/specification/applications/application-registry) page which has similar content for Helm.

The reason for this is we don't want to have to version the non-specification content (everything that's not under the Specification folder), so we want to keep the spec content seperate from the general content. 

I think the current "application package" page content is violating this principle because it currently has spec language.

proposals/quadlet/PATCH_desired-state-quadlet.diff
```diff
@@ -0,0 +1,108 @@
...
++    is_a: DeploymentProfile
++    description: >-
++      Deployment profile for Quadlet-based workloads targeting devices running
++      Podman 5.0 or later with systemd integration. Components are referenced
++      via repository (an oci:// URI) and revision (an OCI tag) consistent with
++      the Helm and Compose OCI publishing model mandated by WG-PROPOSAL-01.
++    rank: 40
++    slot_usage:
++      type:
++        equals_string: "quadlet"
```
matlec, line 23: Why do we use different component type names in the application description (`quadlet.v1`) and desired state document (`quadlet`)?

proposals/quadlet/sup-02-quadlet.md
```diff
@@ -0,0 +1,325 @@
...
+
+---
+
+### Backward compatibility
+
+This SUP is **additive only**.
+
+- The `type` regex expansion from `^(helm\.v3|compose|compose\.v1)$` to `^(helm\.v3|compose|compose\.v1|quadlet\.v1)$` is additive — existing documents continue to validate.
+- `QuadletDeploymentProfile` and `QuadletComponent` are new schema classes — no existing class is modified.
+- Existing WFM implementations that do not support Quadlet will ignore `type: quadlet.v1` deployment profiles per existing spec behavior.
```
matlec: The current `type` slot has a hard regex constraint:

```
...
pattern: ^(helm|compose)$
...
```

A WFM that validates app packages according to the schema would reject an app that uses a `compose.v1` deployment type. I don't see any statement in the current spec that asks to ignore unknown types.

proposals/quadlet/sup-02-quadlet.md
```diff
@@ -0,0 +1,325 @@
...
+````markdown
+## Quadlet Archive Structure
+
+A Quadlet Archive is a gzip-compressed tar archive (`.tar.gz` or `.tgz`) that packages a set of [Podman Quadlet](https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html) unit files for deployment on edge devices running Podman 5.0 or later with systemd integration. The archive MUST conform to the following structural requirements.
+
+### Directory Layout
+
+The archive MUST contain exactly one top-level directory. The name of this directory MUST match the component `name` as specified in the ApplicationDescription.
+
+The top-level directory MAY contain Quadlet unit files according to the [Podman Quadlet specification](https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html). Supported unit file extensions include `.container`, `.network`, `.volume`, `.image`, `.build`, and `.pod`.
```
matlec: Do we need namespacing rules to prevent collisions between deployed Quadlet files? Identically named files from different components or applications (e.g., `app.network`, `db-data.volume`) may all land in the same Quadlet search directory on the host and could overwrite each other.

But, to be honest, that's a general gap today. Margo doesn't currently provide an interoperable way to set a Kubernetes namespace / Helm release name or a Compose project name either. So collision avoidance is left to implementations for Helm and Compose as well. Quadlet is just the most exposed because (unlike Helm and Compose) it has no underlying isolation primitive (to my knowledge) and the files sit flat in a shared directory.

Related:
- https://github.com/margo/specification/issues/142
- https://github.com/margo/specification/issues/183

matlec (commented):
Until the ["Device roles to capabilities" SUP](https://github.com/margo/specification-enhancements/pull/50) lands in the spec, there is no mechanism for a WFM to target Quadlet-capable devices. There is no device role that maps to Quadlets. And even if we map Quadlet support to the `Standalone Device` role, a WFM has no basis to choose between `compose` and `quadlet` if both deployment types are supported by an app package.

Similar to [my comment in the Compose SUP](https://github.com/margo/specification-enhancements/pull/68#discussion_r3411693719), this SUP should define semantics for `wait` and `timeout` for Quadlets. Or deliberately exclude support for these properties with Quadlets.

proposals/quadlet/sup-02-quadlet.md
javatask: .v1 will be removed

javatask:
@matlec, here is my main point about Podman Quadlet and Docker Compose that I'll develop further here:
https://github.com/margo/specification-enhancements/pull/68#issuecomment-4711008166 

It's about the LF-managed project vs the corporate tool that we cannot control or impact.

proposals/quadlet/sup-02-quadlet.md
javatask: Acknowledged.
A dedicated request to add Podman Quadlet to the Margo Feature list was sent

proposals/quadlet/sup-02-quadlet.md
javatask: Acknowledged

proposals/quadlet/sup-02-quadlet.md
javatask: Agreed

proposals/quadlet/sup-02-quadlet.md
javatask: Good catch. Added as a conditional dependency — non-blocking but tracked.
+**Conditional dependency:** If [PR #50](https://github.com/margo/specification-enhancements/pull/50) (remove roles from device capabilities) is approved, it introduces a `supportedDeploymentTypes` array. In that case, `quadlet` MUST be added to that array. This dependency is tracked but non-blocking.

proposals/quadlet/sup-02-quadlet.md
javatask: Agreed, moving section

proposals/quadlet/PATCH_desired-state-quadlet.diff
javatask: Good catch, now it is consistent as ".v1" is removed

proposals/quadlet/sup-02-quadlet.md
javatask: You're rigth.
The correct target regex once both [SUP-01](https://beldenverse.atlassian.net/browse/SUP-01) and this SUP are merged is ^(helm|compose|quadlet)$.

javatask:
@matlec 
> Until the ["Device roles to capabilities" SUP](https://github.com/margo/specification-enhancements/pull/50) lands in the spec, there is no mechanism for a WFM to target Quadlet-capable devices. There is no device role that maps to Quadlets. And even if we map Quadlet support to the `Standalone Device` role, a WFM has no basis to choose between `compose` and `quadlet` if both deployment types are supported by an app package.
> 
> Similar to [my comment in the Compose SUP](https://github.com/margo/specification-enhancements/pull/68#discussion_r3411693719), this SUP should define semantics for `wait` and `timeout` for Quadlets. Or deliberately exclude support for these properties with Quadlets.

1. Runtime selection
For the sake of isolation between SUPs, this SUP does not prescribe changes to the device capabilities manifest or WFM targeting logic. The practical deployment model is: by default, standalone devices use Docker Compose; users (or operators) can manually select the Quadlet deployment profile when deploying to Podman/systemd-based devices. Automated targeting via supportedDeploymentTypes integration is a follow-up coordination item.

2. wait/timeout semantics: We will define them. Proposed mapping:

**`wait`/`timeout` semantics**: We will define them. Proposed mapping:

| Property | Quadlet Semantics |
|----------|-------------------|
| `wait: true` | Device MUST wait until all systemd units generated from the Quadlet archive reach `active (running)` state before reporting deployment success |
| `timeout` | If units have not reached `active (running)` within the specified duration, deployment MUST be reported as failed |
| Unit failure | If any unit enters `failed` state during activation, deployment MUST be reported as failed immediately |
| Health checks | If systemd-level health mechanisms are defined (e.g., `ExecStartPost=` probes, `Type=notify`), implementations SHOULD wait for readiness notification before reporting success |

proposals/quadlet/sup-02-quadlet.md
javatask: Agreed — this is a cross-cutting gap affecting all deployment types. 
This SUP will:

1. Acknowledge namespacing as explicitly **out of scope** with rationale:
   > Namespacing and collision avoidance for Quadlet unit files is out of scope for this SUP. 

2. Add a **RECOMMENDED** (non-normative) mitigation: implementations SHOULD prefix Quadlet unit file names with the application ID or component name when deploying to shared directories (e.g., `myapp-frontend.container` rather than `frontend.container`).

3. Soften the directory name requirement from MUST to SHOULD: "The name of this directory SHOULD match the component `name`" — aligning with the Compose SUP-01 which uses SHOULD for directory naming.

javatask:
@phil-abb @matlec Quadlet SUP was update with your comments included

matlec:
@javatask While implementing the SUP I realized there is a gap in the `pointer` convention of parameters in the Application Description. Today, the spec has a description for Helm (the dot notation for `values.yaml`) and Compose (the plain name of the environment variable). Since Quadlet's parameter injection mechanism is also environment variables (`Environment=KEY=VALUE` in the `[Container]` unit), I'd suggest adopting the same convention as Compose and adding a clause to the [pointer description](https://docs.margo.org/specification/applications/application-description#target-attributes):

> For quadlet deployments, this is the name of the environment variable to set.

I guess this would be sufficient.

javatask:
Hello @pperycz , @lucamaf, @matlec, @MariusKnobbe, please find this SUP as the main for quadlet testing.
[README-margo-community.md](https://github.com/user-attachments/files/29472672/README-margo-community.md)

PR #73, opened 2026-06-19
SUP: Add RFC 9457 common error responses to OpenAPI spec

### LINKED ISSUE #153: Define consistent standard error responses for all 4xx/5xx response codes
### Feature description

Details about standard error responses for all 4xx/5xx - Define one machine-parseable error format (prefer RFC 9457) with stable `type` values, required fields, and retry semantics (`Retry-After`, backoff) so clients can implement consistent UX and automation.​

### Provide adequate technical acceptance criteria(s) associated with this feature below:

- Define standard error responses and implement within all applicable API interactions within the specification

### Although not required, it is highly encouraged to provide feature use-cases below:

- Improved API definitions

### Additional information

Not having error response shapes defined will reduce UX as error responses cannot be interpreted / parsed

vireshnavalli:
Add common error responses to Margo's OpenAPI specification as per RFC9457

### DIFF
--- a/proposals/sup_rfc9457_error_responses.md
+++ b/proposals/sup_rfc9457_error_responses.md
@@ -1,0 +1,137 @@
+# Specification Update Proposal
+
+## Owner
+
+@vireshnavalli
+
+## Summary
+
+Add comprehensive common error responses for 4xx and 5xx HTTP status codes to all API routes following RFC 9457 (Problem Details for HTTP APIs). This standardizes error handling across the margo specification and provides consistent error information to API consumers.
+
+## Reason for proposal
+
+RFC 9457 defines a standard format for problem details in HTTP responses, enabling consistent error handling across REST APIs. Implementing this across all margo API routes provides:
+
+1. **Standardization** - Consistent error response format across all endpoints
+2. **API Consumer Experience** - Clear, predictable error information
+3. **RFC Compliance** - Adherence to industry standard (RFC 9457)
+4. **Better Error Handling** - Enables robust client-side error handling with standard fields
+5. **Specification Clarity** - Addresses specification issue #153
+
+## Requirements alignment acknowledgement
+
+This SUP aligns with margo's commitment to following industry standards and best practices for REST API design. RFC 9457 is an IETF-approved standard for problem details in HTTP APIs, ensuring our specification remains modern and interoperable.
+
+**Related Feature(s):** [Issue #153 - margo/specification](https://github.com/margo/specification/issues/153)
+
+## Technical proposal
+
+### RFC 9457 Error Response Structure
+
+Implement the following standard error response format based on RFC 9457:
+
+```yaml
+ProblemDetails:
+  type: object
+  properties:
+    type:
+      type: string
+      format: uri
+      description: A URI reference identifying the problem type
+      example: "https://api.margo.io/errors/invalid-request"
+    title:
+      type: string
+      description: A short, human-readable summary of the problem
+      example: "Invalid Request"
+    status:
+      type: integer
+      description: The HTTP status code
+      example: 400
+    detail:
+      type: string
+      description: A human-readable explanation of the specific occurrence
+      example: "Missing required field 'name'"
+    instance:
+      type: string
+      format: uri
+      description: A URI reference identifying the specific occurrence
+```
+
+### Common Error Responses
+
+All API routes must include standardized error responses:
+
+#### 4xx Client Errors
+
+- **400 Bad Request**: Invalid request parameters or body
+- **401 Unauthorized**: Missing or invalid authentication
+- **403 Forbidden**: Authenticated but insufficient permissions
+- **404 Not Found**: Requested resource does not exist
+- **409 Conflict**: Request conflicts with current state
+- **422 Unprocessable Entity**: Validation error with semantic meaning
+
+#### 5xx Server Errors
+
+- **500 Internal Server Error**: Unexpected server error
+- **501 Not Implemented**: Feature not yet implemented
+- **503 Service Unavailable**: Server temporarily unable to handle request
+
+### OpenAPI Specification Updates
+
+All endpoints in the OpenAPI specification must include:
+
+```yaml
+responses:
+  '400':
+    description: Bad Request
+    content:
+      application/problem+json:
+        schema:
+          $ref: '#/components/schemas/ProblemDetails'
+  '401':
+    description: Unauthorized
+    content:
+      application/problem+json:
+        schema:
+          $ref: '#/components/schemas/ProblemDetails'
+  '403':
+    description: Forbidden
+    content:
+      application/problem+json:
+        schema:
+          $ref: '#/components/schemas/ProblemDetails'
+  '404':
+    description: Not Found
+    content:
+      application/problem+json:
+        schema:
+          $ref: '#/components/schemas/ProblemDetails'
+  '500':
+    description: Internal Server Error
+    content:
+      application/problem+json:
+        schema:
+          $ref: '#/components/schemas/ProblemDetails'
+  '503':
+    description: Service Unavailable
+    content:
+      application/problem+json:
+        schema:
+          $ref: '#/components/schemas/ProblemDetails'
+```
+
+### Content-Type
+
+Error responses should use the `application/problem+json` media type as specified in RFC 9457.
+
+## Alternatives considered
+
+1. **Custom error response format** - Rejected because it would not conform to industry standards and reduce interoperability with client libraries that expect RFC 9457 format.
+
+2. **Minimal error responses** - Rejected because it provides insufficient error context for API consumers to properly handle and debug issues.
+
+3. **Different error standard (e.g., HAL, JSON:API)** - Rejected in favor of RFC 9457 as it is the most recent IETF standard specifically designed for HTTP APIs.
+
+## Related PRs
+
+- margo/specification PR #188: "feat: adds rfc9457 details to openapi spec"
--- a/proposals/sup_rfc9457_error_responses.md
+++ b/proposals/sup_rfc9457_error_responses.md
@@ -51,7 +51,7 @@ ProblemDetail:
     Extension members (RFC 9457 §3.2) are permitted via additionalProperties.
     Vendors MAY add custom fields alongside standard margo fields.
     Vendor-specific problem types MUST use their own URI namespace —
-    the https://margo.org/problems/ namespace is reserved for this specification.
+    the https://docs.margo.org/specification/problems/ namespace is reserved for this specification.
   required:
     - type
     - title
@@ -68,14 +68,14 @@ ProblemDetail:
 
         For error conditions not defined in this specification, implementations
         MAY define their own type URIs under their own namespace.
-        Implementations MUST NOT use the https://margo.org/problems/ namespace
+        Implementations MUST NOT use the https://docs.margo.org/specification/problems/ namespace
         for custom error types.
 
         Clients MUST use this field for programmatic error handling,
         NOT the status code or title.
         Clients encountering an unknown type URI SHOULD fall back to
         title, detail, and status for display and handling.
-      example: https://margo.org/problems/invalid-digest-header
+      example: https://docs.margo.org/specification/problems/invalid-digest-header
     title:
       type: string
       description: Short human-readable summary of the problem type.
@@ -166,7 +166,7 @@ Type URIs are **permanent stable identifiers and are intentionally unversioned**
 consistent with RFC 9457 design intent and industry practice (Google APIs, Stripe, AWS).
 The URI identifies the error *concept*, not the API version.
 
-- `https://margo.org/problems/invalid-digest-header` means "Missing or invalid content-digest header"
+- `https://docs.margo.org/specification/problems/invalid-digest-header` means "Missing or invalid content-digest header"
   in v1, v2, and all future versions — the concept does not change between API versions
 - If an error concept changes significantly, a **new URI is added** and the old one
   **deprecated** — both remain valid during the transition period
@@ -181,7 +181,7 @@ namespace:
 
 - Vendor URIs MUST use the supplier's own domain
   (e.g. `https://vendor.example.com/problems/sensor-fault`)
-- Vendor URIs MUST NOT use the `https://margo.org/problems/` namespace,
+- Vendor URIs MUST NOT use the `https://docs.margo.org/specification/problems/` namespace,
   which is reserved for this specification
 - Clients encountering an unknown `type` URI SHOULD fall back to using
   `title` and `detail` fields for display, and `status` for HTTP-level handling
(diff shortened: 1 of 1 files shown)


phil-abb:
@margo/technical-wg A new SUP has been started by @vireshnavalli. If anyone is interested in helping work on fleshing out the details of this SUP, please contact @vireshnavalli

proposals/sup_rfc9457_error_responses.md
```diff
@@ -0,0 +1,1150 @@
+# Specification Update Proposal
+
+## Owner
+
+@vireshnavalli
+
+## Summary
+
+Add comprehensive common error responses for 4xx and 5xx HTTP status codes to all API routes
+following RFC 9457 (Problem Details for HTTP APIs). This standardizes error handling across the
```
phil-abb: Add link to RFC 9457 for convenience

proposals/sup_rfc9457_error_responses.md
```diff
@@ -0,0 +1,1150 @@
...
+### Stable Type URI Registry
+
+The following `type` URI values are stable and MUST be used by all implementations.
+Clients MUST use the `type` field for programmatic error handling.
+
+> **Note on URI Dereferenceability:** Per RFC 9457 §4.2, `type` URIs SHOULD be
+> dereferenceable — returning a human-readable page describing the problem type.
+> The URIs listed below currently return 404 as the error catalogue page at
+> `https://margo.org/problems/` has not yet been published. This is a known
+> limitation and a hosted error catalogue is planned as a follow-up documentation
```
phil-abb: Do you mean as a seperate SUP? Or just in a different document as part of this SUP that hasn't been created yet?

proposals/sup_rfc9457_error_responses.md
```diff
@@ -0,0 +1,1150 @@
...
+
+### Stable Type URI Registry
+
+The following `type` URI values are stable and MUST be used by all implementations.
+Clients MUST use the `type` field for programmatic error handling.
+
+> **Note on URI Dereferenceability:** Per RFC 9457 §4.2, `type` URIs SHOULD be
+> dereferenceable — returning a human-readable page describing the problem type.
+> The URIs listed below currently return 404 as the error catalogue page at
+> `https://margo.org/problems/` has not yet been published. This is a known
```
phil-abb: Is `https://margo.org/problems/` normative or just an example?

proposals/sup_rfc9457_error_responses.md
```diff
@@ -0,0 +1,1150 @@
...
+            example: must contain at least one valid role
+  additionalProperties: true
+  # additionalProperties: true allows RFC 9457 §3.5 extension members
+```
+
+---
+
+### Stable Type URI Registry
+
+The following `type` URI values are stable and MUST be used by all implementations.
```
phil-abb, line 153: We may need some more details here, around things like versioning rules and how to handle extensions or depreciations

phil-abb, line 153: Do we need to enable suppliers to define vendor-specific problem URIs?

proposals/sup_rfc9457_error_responses.md
```diff
@@ -0,0 +1,1150 @@
...
+> The URIs listed below currently return 404 as the error catalogue page at
+> `https://margo.org/problems/` has not yet been published. This is a known
+> limitation and a hosted error catalogue is planned as a follow-up documentation
+> task based on this SUP review (Do we really need to host URIs or not). The URIs remain valid stable identifiers regardless of dereferenceability.
+> Clients MUST NOT fetch these URIs at runtime; they are identifiers only.
+
+
+| Type URI | HTTP Status | Description |
+|---|---|---|
+| `https://margo.org/problems/invalid-certificate` | 400 | Certificate is not valid |
```
phil-abb: Keep in mind the approved identity and authentication SUP and [WFM profile SUP](https://github.com/margo/specification-enhancements/blob/main/proposals/wfm-identity-profile.md) specifically will be changing some of this, so we'll want to keep this in mind.

proposals/sup_rfc9457_error_responses.md
```diff
@@ -0,0 +1,1150 @@
...
+
+## Technical Acceptance Criteria
+
+- **AC1**: All error responses MUST use `Content-Type: application/problem+json`
+- **AC2**: `type` field MUST be a stable URI from the Stable Type URI Registry above
+- **AC3**: `type`, `title`, and `status` are REQUIRED fields on all error responses
+- **AC4**: `429` and `503` responses MUST include `Retry-After` response header
+- **AC5**: `500` responses SHOULD include `Retry-After` response header
+- **AC6**: `retryable` field MUST be present on all error responses
+- **AC7**: `backoffStrategy` field MUST be present on all error responses
```
phil-abb: Are we sure we want to make these MUST instead of SHOULDs?

proposals/sup_rfc9457_error_responses.md
```diff
@@ -0,0 +1,1150 @@
...
+- **500 Internal Server Error**: Unexpected server error — SHOULD include `Retry-After` header
+- **501 Not Implemented**: Feature not yet implemented
+- **503 Service Unavailable**: Server temporarily unable to handle request — MUST include
+  `Retry-After` header
+
+---
+
+### OpenAPI Specification Updates
+
+All endpoints in the OpenAPI specification must include applicable error responses using
```
phil-abb: Do we need the full schema here? Can we trim this down to just show some examples instead of having the whole thing?

proposals/sup_rfc9457_error_responses.md
vireshnavalli: @phil-abb thanks for reviewing. I also have same question.
Do we really need to host URIs or not. The URIs remain valid stable identifiers regardless of dereferenceability. Clients MUST NOT fetch these URIs at runtime; they are identifiers only

proposals/sup_rfc9457_error_responses.md
vireshnavalli: added

proposals/sup_rfc9457_error_responses.md
vireshnavalli: We need to make it normative so that all client implementation follow uniform problems URI. Let us know is there any better URI apart from this. Also as per previous comment do we need to host these URIs anywhere to avoid 404 error? because as per RFC URIs should return human-readable page describing the problem type.

proposals/sup_rfc9457_error_responses.md
vireshnavalli: removed whole spec and added few examples

proposals/sup_rfc9457_error_responses.md
vireshnavalli: I will change it to SHOULDs

proposals/sup_rfc9457_error_responses.md
vireshnavalli: Correct me if I am wrong here, this SUP mainly focuses on RFC 9457 and the RFC tells about standardizing the structure of error responses (e.g., type, title, status) rather than dictating API-level versioning or lifecycle management. Because the spec treats the type field as a permanent, dereferenceable URI, managing changes to API errors falls on engineering team. 
Do we need to add margo-specific governance concern regarding versioning and depreciations?

vireshnavalli: We can do that as below. If yes I can update SUP accordingly.

**Current** 
```
type:
  MUST be one of the stable registered URIs
```
---

**Vendor Specific**
```
type:
  MUST be either:
    - A URI from the Stable Type URI Registry (for standard errors), OR
    - A vendor-specific URI 
```

proposals/sup_rfc9457_error_responses.md
phil-abb: If the types are going to be URLs,  the RFC 9457 spec indicates they should be valid direct links to content, but it's not required. It's also my understanding that if these types are valid links, the content is for developers to help understand the problem.

The question is who should own the content: Margo or the WFM supplier. 

To me, it seems like this is content the WFM supplier should own, and they choose to create the content or not. That being said, it seems like it should use `https://<WFM host>/problems/XYZ` instead of `https://margo.org/problems/XYZ`

What do other people think?

proposals/sup_rfc9457_error_responses.md
phil-abb: This relates to the comment right below on who is responsible for creating the content. I'll close this comment

proposals/sup_rfc9457_error_responses.md
phil-abb: I was talking about the type URLs themselves. For example, will "https://margo.org/problems/invalid-certificate" always be "https://margo.org/problems/invalid-certificate" or do we foresee a reason why the content for some of these types might change and become different between version X and Y? So, do the type URLs need to be versioned or not?  

I don't have experience using RFC 9457, so I'm not sure about this. Does anyone have experience with it? If so, is it something that needs to be considered or not?

phil-abb: I think it makes sense to allow WFM suppliers to provide additional types for errors that are not defined by Margo.

vireshnavalli: No — type URIs should NOT be versioned, by design. RFC 9457 §3.1 treats type as a permanent, stable identifier:
```
"type" values are case-sensitive and SHOULD remain stable over time.
The type URI identifies WHAT went wrong, not HOW the API is versioned.
```
The RFC intentionally has no versioning mechanism for type URIs because:
```
type URI = "what category of problem is this"  ← stable concept
API version = "which version of the API"        ← separate concern
```

The only scenarios where a type URI would need to change:
```
Scenario 1: The error concept itself is renamed
  https://margo.org/problems/invalid-certificate
  → renamed to →
  https://margo.org/problems/malformed-certificate
  SOLUTION: Keep old URI, add new URI, deprecate old one
            Both point to same HTTP 400 — clients using old URI still work

Scenario 2: The error splits into two more specific errors
  https://margo.org/problems/invalid-certificate
  → splits into →
  https://margo.org/problems/expired-certificate
  https://margo.org/problems/malformed-certificate
  SOLUTION: Keep old URI as generic fallback, add two new specific ones

Scenario 3: The domain changes (margo.org → margo.io)
  SOLUTION: Redirect old URIs, keep both valid
```

vireshnavalli: added the details for vendor specific problem types and additional parameters.

proposals/sup_rfc9457_error_responses.md
vireshnavalli: changed to shoulds

proposals/sup_rfc9457_error_responses.md
```diff
@@ -0,0 +1,449 @@
...
+| `https://margo.org/problems/signature-verification-failed` | 401 | Payload signature verification failed |
+| `https://margo.org/problems/certificate-not-trusted` | 403 | Client certificate not trusted or revoked |
+| `https://margo.org/problems/not-found` | 404 | Requested resource does not exist |
+| `https://margo.org/problems/conflict` | 409 | Request conflicts with current resource state |
+| `https://margo.org/problems/not-acceptable` | 406 | Server cannot produce requested media type |
+| `https://margo.org/problems/semantic-error` | 422 | Request body contains a semantic error |
+| `https://margo.org/problems/rate-limit-exceeded` | 429 | Rate limit exceeded |
+| `https://margo.org/problems/internal-error` | 500 | Unexpected server error |
+| `https://margo.org/problems/not-implemented` | 501 | Feature not yet implemented |
+| `https://margo.org/problems/service-unavailable` | 503 | Server temporarily unable to handle request |
```
matlec: I think we should refer to the standard [`about:blank` type](https://www.rfc-editor.org/info/rfc9457/#name-aboutblank) for the generic cases and use the Margo namespace only for errors with real Margo-specific meaning.

Errors that (imho) could be simple `about:blank` type errors:
- not-found (404)
- not-acceptable (406)
- conflict (409)
- rate-limit-exceeded (429)
- internal-error (500)
- not-implemented (501)
- service-unavailable (503)

proposals/sup_rfc9457_error_responses.md
matlec: `type` is just an identifier. It has to be a URI, but it does not have to be a http(s) URL. [3.1.1](https://www.rfc-editor.org/info/rfc9457/#section-3.1.1) only raises the "SHOULD resolve" expectation **if** we pick a locator scheme. A `tag:` URI or a URN would also be a perfectly valid type with no hosting obligations at all.

That said, I'd still tend to define `http` URLs under `margo.org` or `docs.margo.org` and commit to publishing a catalogue eventually. The RFC tolerates "interim 404s" :)

proposals/sup_rfc9457_error_responses.md
```diff
@@ -0,0 +1,449 @@
...
+            description: Field path that caused the validation error.
+            example: properties.roles
+          message:
+            type: string
+            description: Human-readable validation error message.
+            example: must contain at least one valid role
+  additionalProperties: true
+  # additionalProperties: true permits RFC 9457 §3.2 extension members.
+  # Margo extension fields: retryable, retryAfterSeconds, backoffStrategy, errors
+  # Vendor extension examples:
```
matlec, line 143: While the examples loosely follow the style we established in the [vendor extensions SUP](https://github.com/margo/specification-enhancements/blob/main/proposals/sup_enable_vendor_extensions.md), they violate the recommendations that [RFC 9457 gives](https://www.rfc-editor.org/info/rfc9457/#name-defining-new-problem-types):

> If such additional members are defined, their names SHOULD start with a letter (ALPHA, as per [[ABNF](https://www.rfc-editor.org/info/rfc9457/#RFC5234)], [Appendix B.1](https://www.rfc-editor.org/info/rfc5234/#appendix-B.1)) and SHOULD comprise characters from ALPHA, DIGIT ([[ABNF](https://www.rfc-editor.org/info/rfc9457/#RFC5234)], [Appendix B.1](https://www.rfc-editor.org/info/rfc5234/#appendix-B.1)), and "_" (so that it can be serialized in formats other than JSON), and they SHOULD be three characters or longer.

Not a big deal, but worth considering if we should change the recommendation to something like `vendor_component` and `vendor_error_code`.

proposals/sup_rfc9457_error_responses.md
```diff
@@ -0,0 +1,449 @@
...
+    type:
+      type: string
+      format: uri
+      description: >
+        URI reference identifying the problem type.
+
+        Margo implementations MUST use a URI from the Stable Type URI Registry
+        (https://margo.org/problems/*).
+
+        Vendor implementations MAY use their own URI namespace
```
matlec: Margo vs vendor implementations reads confusing to me. I'd suggest something like:

> Implementations MUST use the registered Margo type URI for any error condition defined in the Margo specification. For conditions not defined here, implementations MAY define their own type URIs under their own namespace and MUST NOT use the https://margo.org/problems/ namespace.

proposals/sup_rfc9457_error_responses.md
```diff
@@ -0,0 +1,449 @@
...
+### Content-Type
+
+All error responses MUST use the `application/problem+json` media type as specified in RFC 9457.
+Servers MUST NOT return `application/json` for error responses.
+
+---
+
+## Technical Acceptance Criteria
+
+- **AC1**: All error responses SHOULD use `Content-Type: application/problem+json`
```
matlec: This contradicts with line 409:
> All error responses MUST use the `application/problem+json` media type as specified in RFC 9457.

RFC 9457 doesn't mandate the media type, so making it a MUST is a Margo profiling choice. However, imho, it's a reasonable one since clients will dispatch on `content-type`.

vireshnavalli: Yes we will mandate to use  **application/problem+json**, I have corrected AC1 ( changed to MUST)

proposals/sup_rfc9457_error_responses.md
vireshnavalli: Changed this statement.

proposals/sup_rfc9457_error_responses.md
vireshnavalli: Yes makes sense, corrected this.

proposals/sup_rfc9457_error_responses.md
vireshnavalli: Added generic cases also.

proposals/sup_rfc9457_error_responses.md
vireshnavalli: Hi @matlec @phil-abb,
Shall I change `https://margo.org/problems` to  `https://docs.margo.org/specification/problems` which makes sense because when I raise a PR for this SUP on specification repo I will create **problems** folder under **specification** folder with all URIs so that eventually resolve 404s and  we will have catalogue.

phil-abb: Yeah, maybe using `https://docs.margo.org/specification/problems` makes sense if we are going to be hosting these as part of Margo so they can be kept with the specification instead of hosted elsewhere.

phil-abb:
@margo/technical-wg 

This SUP, which defines standard error responses, is ready for final review. Please review the proposal and add any questions/comments to the PR.

phil-abb:
This SUP has been approved so I'm force-merging this PR: https://github.com/margo/specification-enhancements/issues/83

PR #74, opened 2026-06-19
SUP: retain `apiVersion` attribute and remove `kind` attribute in ApplicationDescription

### LINKED ISSUE #134: Document expected `apiVersion` values for all API resources
### Bug description

The spec uses `apiVersion` as a required field in request bodies (e.g., [Device Client Onboarding](https://docs.margo.org/specification/margo-management-interface/device-client-onboarding)), but the table only states:

> `apiVersion` - Identifier of the version the API resource follows.

This doesn't specify what the actual value should be. The example shows `"onboarding.margo.org/v1alpha1"`, but this is never explicitly documented as the expected value.

**Issues:**

1. Implementers don't know what value to validate against
2. Clients don't know what value to send
3. The pattern/convention for `apiVersion` strings is undefined (e.g., `{resource}.margo.org/{version}`)

### Proposed fix

For each API resource, explicitly document:
- The expected `apiVersion` value (e.g., `onboarding.margo.org/v1alpha1`)
- Whether servers should reject unknown versions or be lenient
- The versioning convention (e.g., `v1alpha1` --> `v1beta1` --> `v1`)

### Anything else (optional)

_No response_

vireshnavalli:
Margo's OpenAPIs have apiversion information explicitly embedded within each API manifest. It is not required as version information can be found from API path itself. So this SUP will remove apiversion and kind fields from all API manifests. And retain apiVersion attribute and remove kind attribute in ApplicationDescription

### DIFF
--- a/proposals/sup_apiversion_app_description_deployment.md
+++ b/proposals/sup_apiversion_app_description_deployment.md
@@ -1,0 +1,73 @@
+# Specification Update Proposal
+
+## Owner
+
+@vireshnavalli
+
+## Summary
+
+Retain the `apiVersion` attribute in `ApplicationDescription` and `ApplicationDeployment` structures. Although API routes include the version within the URL path, keeping the apiVersion in these structures provides explicit version information and ensures consistency with semantic versioning requirements.
+
+## Reason for proposal
+
+As API routes include version within URL, there is debate about the necessity of explicitly having 'apiVersion' attribute for margo specified REST APIs. However, maintaining the apiVersion in ApplicationDescription and ApplicationDeployment provides:
+
+1. Explicit version tracking at the application definition level
+2. Clear semantic versioning compliance
+3. Version information available without parsing URL paths
+4. Better consistency across specification structures
+5. Addresses specification issue #134
+
+## Requirements alignment acknowledgement
+
+This SUP aligns with the margo specification requirements for clear API versioning and semantic versioning practices. The apiVersion attribute ensures that application descriptions and deployments carry explicit version information, improving clarity and consistency across the specification.
+
+**Related Feature(s):** [Issue #134 - margo/specification](https://github.com/margo/specification/issues/134)
+
+## Technical proposal
+
+### Changes to ApplicationDescription
+
+Retain the `apiVersion` field in the `ApplicationDescription` schema:
+
+```yaml
+ApplicationDescription:
+  type: object
+  properties:
+    apiVersion:
+      type: string
+      description: Semantic version of the application API specification
+      example: "1.0.0"
+    # ... other properties
+```
+
+### Changes to ApplicationDeployment
+
+Retain the `apiVersion` field in the `ApplicationDeployment` schema:
+
+```yaml
+ApplicationDeployment:
+  type: object
+  properties:
+    apiVersion:
+      type: string
+      description: Semantic version of the application API specification
+      example: "1.0.0"
+    # ... other properties
+```
+
+### Version Consistency
+
+- The `apiVersion` value must match the OpenAPI specification version
+- Must follow semantic versioning format (MAJOR.MINOR.PATCH)
+- Should be maintained in sync with API route versioning
+
+## Alternatives considered
+
+1. **Remove apiVersion entirely** - Rejected because it removes explicit version information from application definitions, making it harder to track versioning at the application level without parsing URL routes.
+
+2. **Include version only in API routes** - Rejected because it provides insufficient version metadata in application descriptors and deployment configurations.
+
+## Related PRs
+
+- margo/specification PR #189: "chore: retains apiversion in app description and app deployments"
--- a/proposals/sup_apiversion_app_description_deployment.md
+++ b/proposals/sup_apiversion_app_description_deployment.md
@@ -64,9 +64,8 @@ ApplicationDeployment:
 
 ## Alternatives considered
 
-1. **Remove apiVersion entirely** - Rejected because it removes explicit version information from application definitions, making it harder to track versioning at the application level without parsing URL routes.
+1. **Remove apiVersion entirely** - Rejected because it removes explicit version information from application definitions (application description) and application deployment manifest, making it harder to track versioning at the application description/deployment manifests. Remove apiVersion from the body of the API requests as it is part of the API route, keep the apiVersion in ApplicationDescription (and ApplicationDeployment)
 
-2. **Include version only in API routes** - Rejected because it provides insufficient version metadata in application descriptors and deployment configurations.
 
 ## Related PRs
 
--- a/proposals/sup_apiversion_app_description_deployment.md
+++ b/proposals/sup_apiversion_app_description_deployment.md
@@ -40,8 +40,8 @@ It is an independent document  version identifier.
 
 | Document | apiVersion value |
 |---|---|
-| `ApplicationDescription` | `margo.org/v1alpha1` |
-| `ApplicationDeployment` | `application.margo.org/v1alpha1` |
+| `ApplicationDescription` | `v1alpha1` |
+| `ApplicationDeployment` | `v1alpha1` |
 
 These values are **independent** of:
 - OpenAPI spec version (`1.0.0`)
@@ -65,8 +65,8 @@ ApplicationDescription:
         This is independent of the OpenAPI specification version and API route version.
         MUST be one of the Margo defined stable value.
       enum:
-        - margo.org/v1alpha1
-      example: margo.org/v1alpha1
+        - v1alpha1
+      example: v1alpha1
     kind:
       type: string
       enum: [ApplicationDescription]
@@ -91,8 +91,8 @@ ApplicationDeployment:
         This is independent of the OpenAPI specification version and API route version.
         MUST be one of the Margo defined stable value.
       enum:
-        - application.margo.org/v1alpha1
-      example: application.margo.org/v1alpha1
+        - v1alpha1
+      example: v1alpha1
     kind:
       type: string
       enum: [ApplicationDeployment]
@@ -108,9 +108,9 @@ The progression path for breaking changes is:
 
 | Stage | ApplicationDescription | ApplicationDeployment |
 |---|---|---|
-| Current | `margo.org/v1alpha1` | `application.margo.org/v1alpha1` |
-| Next breaking change | `margo.org/v1alpha2` | `application.margo.org/v1alpha2` |
-| Stable release | `margo.org/v1` | `application.margo.org/v1` |
+| Current | `v1alpha1` | `v1alpha1` |
+| Next breaking change | `v1alpha2` | `v1alpha2` |
+| Stable release | `v1` | `v1` |
 
 **Rules:**
 
@@ -130,10 +130,10 @@ The progression path for breaking changes is:
 ### Version Independence Clarification
 
 ```
-OpenAPI spec version:   1.0.0                          ← spec metadata, semver
-API route version:      /api/v1/                       ← structure contract version via URL path segment
-ApplicationDescription: margo.org/v1alpha1            ← structure contract version via document
-ApplicationDeployment:  application.margo.org/v1alpha1 ← structure contract version via document
+OpenAPI spec version:   1.0.0    ← spec metadata, semver
+API route version:      /api/v1/ ← structure contract version via URL path segment
+ApplicationDescription: v1alpha1 ← structure contract version via document
+ApplicationDeployment:  v1alpha1 ← structure contract version via document
 
 These are three independent versioning axes.
 A change to the API route version does NOT require a change to apiVersion.
@@ -156,7 +156,7 @@ A change to apiVersion does NOT require a change to the API route version.
     evolution and API route evolution are independent.
 
 4. **Use semver (1.0.0) for apiVersion** — Rejected in favour of Kubernetes-style API group
-   versioning (`margo.org/v1-alpha1`) which is more expressive about stability stage
+   versioning (`v1alpha1`) which is more expressive about stability stage
    (alpha/beta/stable) and is familiar to edge/cloud-native implementors.
 
 ## Related PRs
--- a/proposals/sup_apiversion_app_description_deployment.md
+++ b/proposals/sup_apiversion_app_description_deployment.md
@@ -7,27 +7,24 @@
 ## Summary
 
 Retain the `apiVersion` attribute in `ApplicationDescription` and `ApplicationDeployment`
-structures. The `apiVersion` in these structures is independent of the Margo's OpenAPI specification
-version and API route versioning — it identifies the  version of the document itself,
-following a Kubernetes-style API group versioning convention.
+structures. The `apiVersion` in these structures is independent of the Margo's OpenAPI specification — it identifies the version of the contract the document satisfies.
 
 ## Reason for proposal
 
-As API routes include version within the URL path, there is debate about the necessity of
-explicitly having an `apiVersion` attribute. However, the `apiVersion` in `ApplicationDescription`
-and `ApplicationDeployment` serves a distinct purpose — it identifies the ** version of the
-document**, not the API route version. Maintaining it provides:
+As API routes include structure contract versions within the URL path, there is debate about the necessity of
+explicitly having an `apiVersion` attribute. Since there are no URL endpoints for the `ApplicationDescription` and `ApplicationDeployment` structures, the `apiVersion` serves to identify the ** structure contract version of the
+document satisfies**, not the OpenAPI specification version. Maintaining it provides:
 
-1. Explicit  version tracking at the document level — independent of API route versioning
-2. Clear distinction between API route version (`/api/v1/`) and document  version
-3. Version available without parsing URL paths or OpenAPI spec metadata
+1. Explicit version tracking at the document level — since there is no URL endpoint to indicate the structure contract version.
+2. Clear distinction between specification version (1.0.0) and structure contract version (v1, v2, v2alpha1)
+3. The structure's contract version can be determined directly from the document since no URL is available
 4. Alignment with Kubernetes-style API group versioning conventions
-5. Enables document evolution independently of API route changes
+5. Enables document evolution independently of API and other specification changes
 6. Addresses specification [Issue #134 - margo/specification](https://github.com/margo/specification/issues/134)
 
 ## Requirements alignment acknowledgement
 
-This SUP aligns with the margo specification requirements for clear document  versioning.
+This SUP aligns with the Margo specification requirements for clear contract versioning.
 The `apiVersion` attribute in `ApplicationDescription` and `ApplicationDeployment` is explicitly
 **not** tied to:
 - The OpenAPI specification version (e.g. `1.0.0`)
@@ -43,7 +40,7 @@ It is an independent document  version identifier.
 
 | Document | apiVersion value |
 |---|---|
-| `ApplicationDescription` | `margo.org/v1-alpha1` |
+| `ApplicationDescription` | `margo.org/v1alpha1` |
 | `ApplicationDeployment` | `application.margo.org/v1alpha1` |
 
 These values are **independent** of:
@@ -68,8 +65,8 @@ ApplicationDescription:
         This is independent of the OpenAPI specification version and API route version.
         MUST be one of the Margo defined stable value.
       enum:
-        - margo.org/v1-alpha1
-      example: margo.org/v1-alpha1
+        - margo.org/v1alpha1
+      example: margo.org/v1alpha1
     kind:
       type: string
       enum: [ApplicationDescription]
@@ -111,8 +108,8 @@ The progression path for breaking changes is:
 
 | Stage | ApplicationDescription | ApplicationDeployment |
 |---|---|---|
-| Current | `margo.org/v1-alpha1` | `application.margo.org/v1alpha1` |
-| Next breaking change | `margo.org/v1-alpha2` | `application.margo.org/v1alpha2` |
+| Current | `margo.org/v1alpha1` | `application.margo.org/v1alpha1` |
+| Next breaking change | `margo.org/v1alpha2` | `application.margo.org/v1alpha2` |
 | Stable release | `margo.org/v1` | `application.margo.org/v1` |
 
 **Rules:**
@@ -124,7 +121,7 @@ The progression path for breaking changes is:
 - Both old and new `apiVersion` values SHOULD be supported simultaneously during a
   transition period of at least one major specification release
 
-> **Note:** The versioning path (e.g. `v1-alpha1` → `v1-alpha2`) is not yet fully defined.
+> **Note:** The versioning path (e.g. `v1alpha1` → `v1alpha2`) is not yet fully defined.
 > The exact criteria for what constitutes a "significantly breaking change" requiring a new
 > `apiVersion` is an open question to be resolved during SUP review.
 
@@ -134,9 +131,9 @@ The progression path for breaking changes is:
 
 ```
 OpenAPI spec version:   1.0.0                          ← spec metadata, semver
-API route version:      /api/v1/                       ← URL path segment
-ApplicationDescription: margo.org/v1-alpha1            ← document  version
-ApplicationDeployment:  application.margo.org/v1alpha1 ← document  version
+API route version:      /api/v1/                       ← structure contract version via URL path segment
+ApplicationDescription: margo.org/v1alpha1            ← structure contract version via document
+ApplicationDeployment:  application.margo.org/v1alpha1 ← structure contract version via document
 
 These are three independent versioning axes.
 A change to the API route version does NOT require a change to apiVersion.
--- a/proposals/sup_apiversion_app_description_deployment.md
+++ b/proposals/sup_apiversion_app_description_deployment.md
@@ -7,20 +7,18 @@
 ## Summary
 
 Retain the `apiVersion` attribute in `ApplicationDescription` and `ApplicationDeployment`
-structures. The `apiVersion` in these structures is independent of the Margo's OpenAPI specification — it identifies the version of the contract the document satisfies.
+structures. The `apiVersion` in these structures is independent of the Margo's OpenAPI specification — it identifies the version of the contract the document satisfies. Also need to remove `apiVersion` and `kind` attributes from all API routes of Margo's OpenAPI specification except `ApplicationDeployment` as per this SUP's rationale.
 
 ## Reason for proposal
 
 As API routes include structure contract versions within the URL path, there is debate about the necessity of
-explicitly having an `apiVersion` attribute. Since there are no URL endpoints for the `ApplicationDescription` and `ApplicationDeployment` structures, the `apiVersion` serves to identify the ** structure contract version of the
-document satisfies**, not the OpenAPI specification version. Maintaining it provides:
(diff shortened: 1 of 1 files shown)


phil-abb:
@margo/technical-wg A new SUP has been started by @vireshnavalli. If anyone is interested in helping work on fleshing out the details of this SUP, please contact @vireshnavalli

phil-abb:
I think we're mixing up "spec version" with "contract version." The "apiVersion" in these documents should reflect the contract version of the "api" and not reflect the latest Margo specification version. Contract versions are typically done with just the major version (v1, v2) and potentially pre-release labels (v1beta1, v2alpha1) and only change when there is a breaking change to that specific contract. Contract versions and semantic versions for product releases are not the same thing. 

So, just because there is a v2.0.0 release of the Margo specification doesn't mean the contract version of the application description or application deployment has changed. If version 2.0.0 doesn't introduce a "breaking" change to either of these documents, there is no reason to change the "apiVersion" to v2 just because that's the version of the latest Margo specification. We shouldn't be forcing change to be made merely to update a version number. 

If we try to align the apiVersion with the semantic version of each specification release:
1. This forces code changes for reasons that have nothing to do with why the code needs to change
2. It confuses compatibility
3. It means the api implementations have to track all these different released specification versions for reasons that are not relevant
4. It makes backward compatibility guarantees more difficult to manage

Also, not every contract change is a breaking change. You can make additive changes to a contract and not break the existing functionality. The new contract version should only be introduced when there are actual breaking changes.

proposals/sup_apiversion_app_description_deployment.md
```diff
@@ -0,0 +1,167 @@
...
+## Owner
+
+@vireshnavalli
+
+## Summary
+
+Retain the `apiVersion` attribute in `ApplicationDescription` and `ApplicationDeployment`
+structures. The `apiVersion` in these structures is independent of the Margo's OpenAPI specification
+version and API route versioning — it identifies the  version of the document itself,
+following a Kubernetes-style API group versioning convention.
```
phil-abb:
```suggestion
Retain the `apiVersion` attribute in `ApplicationDescription` and `ApplicationDeployment`
structures. The `apiVersion` in these structures is independent of the Margo's OpenAPI specification — it identifies the version of the contract the document satisfies.
```

proposals/sup_apiversion_app_description_deployment.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Reason for proposal
+
+As API routes include version within the URL path, there is debate about the necessity of
+explicitly having an `apiVersion` attribute. However, the `apiVersion` in `ApplicationDescription`
+and `ApplicationDeployment` serves a distinct purpose — it identifies the ** version of the
+document**, not the API route version. Maintaining it provides:
```
phil-abb:
```suggestion
As API routes include structure contract versions within the URL path, there is debate about the necessity of
explicitly having an `apiVersion` attribute. Since there are no URL endpoints for the `ApplicationDescription` and `ApplicationDeployment` structures, the `apiVersion` serves to identify the ** structure contract version of the
document satisfies**, not the OpenAPI specification version. Maintaining it provides:
```

proposals/sup_apiversion_app_description_deployment.md
```diff
@@ -0,0 +1,167 @@
...
+
+1. Explicit  version tracking at the document level — independent of API route versioning
+2. Clear distinction between API route version (`/api/v1/`) and document  version
+3. Version available without parsing URL paths or OpenAPI spec metadata
+4. Alignment with Kubernetes-style API group versioning conventions
+5. Enables document evolution independently of API route changes
+6. Addresses specification [Issue #134 - margo/specification](https://github.com/margo/specification/issues/134)
```
phil-abb:
```suggestion
1. Explicit version tracking at the document level — since there is no URL endpoint to indicate the structure contract version.
2. Clear distinction between specification version (1.0.0) and structure contract version (v1, v2, v2alpha1)
3. The structure's contract version can be determined directly from the document since no URL is available
4. Alignment with Kubernetes-style API group versioning conventions
5. Enables document evolution independently of API and other specification changes
6. Addresses specification [Issue #134 - margo/specification](https://github.com/margo/specification/issues/134)
```

proposals/sup_apiversion_app_description_deployment.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Requirements alignment acknowledgement
+
+This SUP aligns with the margo specification requirements for clear document  versioning.
```
phil-abb:
```suggestion
This SUP aligns with the Margo specification requirements for clear contract versioning.
```

proposals/sup_apiversion_app_description_deployment.md
```diff
@@ -0,0 +1,167 @@
...
+**Related Feature(s):** [Issue #134 - margo/specification](https://github.com/margo/specification/issues/134)
+
+## Technical proposal
+
+### apiVersion Values
+
+| Document | apiVersion value |
+|---|---|
+| `ApplicationDescription` | `margo.org/v1-alpha1` |
+| `ApplicationDeployment` | `application.margo.org/v1alpha1` |
```
phil-abb: I'd propose we get rid of the "margo.org" and "application.margo.org" prefixes and just stick to the major version (v1, v2, v3)

I'll comment later about pre-release labels (v1alpha1, v2beta1)

We're not creating CRDs, so I don't think we need to try to mimic what Kubernetes has just because that's how it's done in CRDs.

proposals/sup_apiversion_app_description_deployment.md
```diff
@@ -0,0 +1,167 @@
...
+---
+
+### Version Progression
+
+The `apiVersion` follows a Kubernetes-style API group versioning convention.
+The progression path for breaking changes is:
+
+| Stage | ApplicationDescription | ApplicationDeployment |
+|---|---|---|
+| Current | `margo.org/v1-alpha1` | `application.margo.org/v1alpha1` |
```
phil-abb: If we're going to use prerelease labels, it shouldn't have a dash, just v1alpha1, v2beta1, etc.

proposals/sup_apiversion_app_description.md
```diff
@@ -0,0 +1,167 @@
...
+
+| Stage | ApplicationDescription | ApplicationDeployment |
+|---|---|---|
+| Current | `margo.org/v1-alpha1` | `application.margo.org/v1alpha1` |
+| Next breaking change | `margo.org/v1-alpha2` | `application.margo.org/v1alpha2` |
+| Stable release | `margo.org/v1` | `application.margo.org/v1` |
+
+**Rules:**
+
+- A new `apiVersion` value MUST be introduced for any **significantly breaking**  change
```
phil-abb, line 92: We should enumerate what we think are breaking changes and what we think are non-breaking changes:

For example:

Breaking changes:
- Renaming an endpoint, properties/fields, enumeration value, or parameter
- Removing an endpoint, properties/fields, enumeration value, or parameter
- Changing data types or expected format
- Making optional things required
- Changing the semantics or behavior of an endpoint
- Changing HTTP method or response codes
- Making any validations stricter
- Changing authentication/authorization rules

Non-Breaking changes:
- Adding new endpoints, properties/fields, or parameters
- Adding optional parameters
- Adding headers
- Adding metadata
- Fixing incorrect behaviors (bugs)

Gray areas - can be breaking if not handled well:
- Adding enumeration values
- Changing defaults
- Reordering fields
- Changing error responses

proposals/sup_apiversion_app_description.md
```diff
@@ -0,0 +1,167 @@
...
+- Non-breaking additive changes (new optional fields) MAY be made within the same `apiVersion`
+- Clients MUST reject documents with an unrecognised `apiVersion`
+- Servers MUST NOT serve documents with a deprecated `apiVersion` after its removal date
+- Both old and new `apiVersion` values SHOULD be supported simultaneously during a
+  transition period of at least one major specification release
```
phil-abb, line 97: I'm going to have a follow-up comment about these topics on the PR itself and not inline

proposals/sup_apiversion_app_description_deployment.md
```diff
@@ -0,0 +1,167 @@
...
+
+**Related Feature(s):** [Issue #134 - margo/specification](https://github.com/margo/specification/issues/134)
+
+## Technical proposal
+
+### apiVersion Values
+
+| Document | apiVersion value |
+|---|---|
+| `ApplicationDescription` | `margo.org/v1-alpha1` |
```
phil-abb:
```suggestion
| `ApplicationDescription` | `margo.org/v1alpha1` |
```

proposals/sup_apiversion_app_description_deployment.md
```diff
@@ -0,0 +1,167 @@
...
+  required: [apiVersion, kind]
+  properties:
+    apiVersion:
+      type: string
+      description: >
+         version of the ApplicationDescription document.
+        This is independent of the OpenAPI specification version and API route version.
+        MUST be one of the Margo defined stable value.
+      enum:
+        - margo.org/v1-alpha1
```
phil-abb:
```suggestion
        - margo.org/v1alpha1
```

proposals/sup_apiversion_app_description_deployment.md
```diff
@@ -0,0 +1,167 @@
...
+      example: margo.org/v1-alpha1
```
phil-abb:
```suggestion
      example: margo.org/v1alpha1
```

proposals/sup_apiversion_app_description_deployment.md
phil-abb:
```suggestion
| Current | `margo.org/v1alpha1` | `application.margo.org/v1alpha1` |
```

proposals/sup_apiversion_app_description_deployment.md
```diff
@@ -0,0 +1,167 @@
...
+
+### Version Progression
+
+The `apiVersion` follows a Kubernetes-style API group versioning convention.
+The progression path for breaking changes is:
+
+| Stage | ApplicationDescription | ApplicationDeployment |
+|---|---|---|
+| Current | `margo.org/v1-alpha1` | `application.margo.org/v1alpha1` |
+| Next breaking change | `margo.org/v1-alpha2` | `application.margo.org/v1alpha2` |
```
phil-abb:
```suggestion
| Next breaking change | `margo.org/v1alpha2` | `application.margo.org/v1alpha2` |
```

proposals/sup_apiversion_app_description_deployment.md
```diff
@@ -0,0 +1,167 @@
...
+**Rules:**
+
+- A new `apiVersion` value MUST be introduced for any **significantly breaking**  change
+- Non-breaking additive changes (new optional fields) MAY be made within the same `apiVersion`
+- Clients MUST reject documents with an unrecognised `apiVersion`
+- Servers MUST NOT serve documents with a deprecated `apiVersion` after its removal date
+- Both old and new `apiVersion` values SHOULD be supported simultaneously during a
+  transition period of at least one major specification release
+
+> **Note:** The versioning path (e.g. `v1-alpha1` → `v1-alpha2`) is not yet fully defined.
```
phil-abb:
```suggestion
> **Note:** The versioning path (e.g. `v1alpha1` → `v1alpha2`) is not yet fully defined.
```

proposals/sup_apiversion_app_description_deployment.md
```diff
@@ -0,0 +1,167 @@
...
+
+---
+
+### Version Independence Clarification
+
+```
+OpenAPI spec version:   1.0.0                          ← spec metadata, semver
+API route version:      /api/v1/                       ← URL path segment
+ApplicationDescription: margo.org/v1-alpha1            ← document  version
+ApplicationDeployment:  application.margo.org/v1alpha1 ← document  version
```
phil-abb:
```suggestion
OpenAPI spec version:   1.0.0                          ← spec metadata, semver
API route version:      /api/v1/                       ← structure contract version via URL path segment
ApplicationDescription: margo.org/v1alpha1            ← structure contract version via document
ApplicationDeployment:  application.margo.org/v1alpha1 ← structure contract version via document
```

phil-abb:
Regarding the use of pre-release labels (v1alpha2, v2beta1, etc.)

- I propose we do not use these until we need to use them. Since the Margo specification has not been released officially, we stick with just "v1" for now, without "alpha1", "beta1", etc. being part of the version. 
- I propose we only add the "alpha1" and " beta1" parts to the version if we have an official release of the specification that includes a pre-release contract version that is intentionally being released before it is finalized. This is not an uncommon practice.
- I propose we do not use the pre-release labels just for internal use.

phil-abb:
I see an opportunity here to define our stance/rules for backwards compatibility. I think it's a bit more than what the original issue covers, but I think it's very applicable to this SUP because this SUP is defining how contract versioning will be supported.

I see two areas where we can provide guidelines in this SUP:

1. Deals with how we handle contract versioning in both the URL paths and these two documents. The SUP touches on this already, but I think it would be good to make it clear that the contract versions are not the same as the Margo Specification/OpenAPI version. The version in the URL paths and these documents reflects the structure's contract version, which means these versions change independently. When there is a breaking change to any of these structures, the contract increases by 1 major version regardless of the Margo specification version. 

For example:

- We release Margo specification version 1.0.0 with "/v1/deployments" and "/v1/capabilities"
- We release Margo specification version 2.0.0 with a new deployments contract, so we have endpoints "/v1/deployments", "/v2/deployments", and "/v1/capabilities"
- We release Margo specification version 3.0.0 with a new capabilities contract, so we have endpoints "/v1/deployments", "v2/deployments", "/v1/capabilities", and "/v2/capabilities". It does not use "/v3/capabilities" just because the version of the specification is 3.0.0

2. Deals with our backward compatibility rules. As the Margo specification evolves, there will undoubtedly be changes that require our contracts to change. We cannot expect everything to be updated when a supplier chooses to adopt the latest version of the Margo specification, so there needs to be some concessions for backwards compatibility. We need to determine our rules for how long older versions need to continue to be supported.  Supporting all older Margo specification versions forever seems excessive. Supporting only the current and previous Margo specification versions doesn't seem to be enough.

proposals/sup_apiversion_app_description_deployment.md
vireshnavalli: Yes sounds good, I will keep **v1alpha1** for now. lets hear what others say @margo/technical-wg

vireshnavalli:
> Regarding the use of pre-release labels (v1alpha2, v2beta1, etc.)
> 
> * I propose we do not use these until we need to use them. Since the Margo specification has not been released officially, we stick with just "v1" for now, without "alpha1", "beta1", etc. being part of the version.
> * I propose we only add the "alpha1" and " beta1" parts to the version if we have an official release of the specification that includes a pre-release contract version that is intentionally being released before it is finalized. This is not an uncommon practice.
> * I propose we do not use the pre-release labels just for internal use.

Yes sounds good, I will keep v1alpha1 for now. lets hear what others say @margo/technical-wg

vireshnavalli:
> I see an opportunity here to define our stance/rules for backwards compatibility. I think it's a bit more than what the original issue covers, but I think it's very applicable to this SUP because this SUP is defining how contract versioning will be supported.
> 
> I see two areas where we can provide guidelines in this SUP:
> 
> 1. Deals with how we handle contract versioning in both the URL paths and these two documents. The SUP touches on this already, but I think it would be good to make it clear that the contract versions are not the same as the Margo Specification/OpenAPI version. The version in the URL paths and these documents reflects the structure's contract version, which means these versions change independently. When there is a breaking change to any of these structures, the contract increases by 1 major version regardless of the Margo specification version.
> 
> For example:
> 
> * We release Margo specification version 1.0.0 with "/v1/deployments" and "/v1/capabilities"
> * We release Margo specification version 2.0.0 with a new deployments contract, so we have endpoints "/v1/deployments", "/v2/deployments", and "/v1/capabilities"
> * We release Margo specification version 3.0.0 with a new capabilities contract, so we have endpoints "/v1/deployments", "v2/deployments", "/v1/capabilities", and "/v2/capabilities". It does not use "/v3/capabilities" just because the version of the specification is 3.0.0
> 
> 2. Deals with our backward compatibility rules. As the Margo specification evolves, there will undoubtedly be changes that require our contracts to change. We cannot expect everything to be updated when a supplier chooses to adopt the latest version of the Margo specification, so there needs to be some concessions for backwards compatibility. We need to determine our rules for how long older versions need to continue to be supported.  Supporting all older Margo specification versions forever seems excessive. Supporting only the current and previous Margo specification versions doesn't seem to be enough.

Thanks for suggestion @phil-abb , I will cover these points in new updates to this SUP.

proposals/sup_apiversion_app_description_deployment.md
phil-abb: The dash was removed with the latest changes so I'll close this comment

proposals/sup_apiversion_app_description.md
phil-abb: I did this so I'm closing this comment

proposals/sup_apiversion_app_description_deployment.md
phil-abb: I'm resolving this comment since there's a follow-up post directly on the PR.

phil-abb:
> Thanks for suggestion @phil-abb , I will cover these points in new updates to this SUP.

Here's an example for item number 2 for backwards compatibility:

- Margo guarantees backward compatibility at the wire/schema level across all versions within a major version.
- Implementations should support at least the current version plus the previous # major versions.
- Deprecated features will be retained for a minimum of # years before removal.
- All payloads must be forward-compatible via tolerant parsing (ignore unknown fields).
- Version negotiation or capability declaration should be used where interoperability is required across versions.

I'm not sure what is reasonable for how many years to support deprecated APIs. 5-7 years maybe? This feels like a long time, but I'm more used to application APIs and not OT use cases.

Also, the number of previous versions to support depends on how frequently things end up changing, so it may be more of a rule for how long to support deprecated APIs instead of how many previous versions.

proposals/sup_apiversion_app_description.md
vireshnavalli: Included in SUP.

matlec (commented):
Hey @vireshnavalli!

These are my thoughts on the SUP:
1. The SUP only addresses how `apiVersion` is retained in `ApplicationDescription` and `ApplicationDeployment` (see (4)) but we never spell out the plan to actually remove `apiVersion` from the payloads (`OnboardingRequest`, ...).
2. I agree with @phil-abb on the idea of using `v1` (instead of `v1alpha1` etc.) while we are pre-draft.
3. In [the original issue](https://github.com/margo/specification/issues/134#issuecomment-4136063260) we treated `kind` and `apiVersion` as a pair. The route is enough to identify the payload type, so `kind` is redundant wherever `apiVersion` is. If we keep either, let's be deliberate.
4. The SUP's rationale says that there are "no URL endpoints for `ApplicationDescription` and `ApplicationDeployment`", but the spec has an endpoint for retrieving the `ApplicationDeployment` (`GET /api/v1/clients/{clientId}/deployments/{deploymentId}/{digest}`). So `ApplicationDeployment` **is** served over a versioned route, and by the same reasoning in (3) its `apiVersion` is redundant. If we keep `apiVersion` in `ApplicationDeployment`, the reason can't be "no endpoint".

phil-abb:
The SUP's rationale says that there are "no URL endpoints for ApplicationDescription and ApplicationDeployment", but the spec has an endpoint for retrieving the ApplicationDeployment (GET /api/v1/clients/{clientId}/deployments/{deploymentId}/{digest}). So ApplicationDeployment is served over a versioned route, and by the same reasoning in (3) its apiVersion is redundant. If we keep apiVersion in ApplicationDeployment, the reason can't be "no endpoint".

Yeah, you're right. I keep thinking back to the old idea of using OCI for this. Since there are endpoints, I don't think we need the "apiVersion" in the Application Deployment manifest.


Philip Presson
Digital Platform Architect for Service Applications
***@***.***?anonymous&ep=bwmEmailSignature>
Book time to meet with ***@***.***?anonymous&ep=bwmEmailSignature>


________________________________

vireshnavalli:
> These are my thoughts on the SUP:
> 
> 1. The SUP only addresses how `apiVersion` is retained in `ApplicationDescription` and `ApplicationDeployment` (see (4)) but we never spell out the plan to actually remove `apiVersion` from the payloads (`OnboardingRequest`, ...).
> 2. I agree with @phil-abb on the idea of using `v1` (instead of `v1alpha1` etc.) while we are pre-draft.
> 3. In [the original issue](https://github.com/margo/specification/issues/134#issuecomment-4136063260) we treated `kind` and `apiVersion` as a pair. The route is enough to identify the payload type, so `kind` is redundant wherever `apiVersion` is. If we keep either, let's be deliberate.
> 4. The SUP's rationale says that there are "no URL endpoints for `ApplicationDescription` and `ApplicationDeployment`", but the spec has an endpoint for retrieving the `ApplicationDeployment` (`GET /api/v1/clients/{clientId}/deployments/{deploymentId}/{digest}`). So `ApplicationDeployment` **is** served over a versioned route, and by the same reasoning in (3) its `apiVersion` is redundant. If we keep `apiVersion` in `ApplicationDeployment`, the reason can't be "no endpoint".

Thanks @matlec @phil-abb for review
I will correct #.1 #.2 #.3 
Regarding the rationale of the SUP we can correct it as below. What you people say?
Although `ApplicationDeployment` is served over a versioned route
(`GET /api/v1/clients/{clientId}/deployments/{deploymentId}/{digest}`),
`apiVersion` is retained because: it has lifecycle outside the API scope of the API payloads and may reside either in the OCI registry or in the WFM object store, hence it needs an explicit version

phil-abb:
> Thanks @matlec @phil-abb for review I will correct #.1 #.2 #.3 Regarding the rationale of the SUP we can correct it as below. What you people say? Although `ApplicationDeployment` is served over a versioned route (`GET /api/v1/clients/{clientId}/deployments/{deploymentId}/{digest}`), `apiVersion` is retained because: it has lifecycle outside the API scope of the API payloads and may reside either in the OCI registry or in the WFM object store, hence it needs an explicit version

Fair point. Clearly, I don't have a strong opinion on this because I keep going back and forth 🧇 😄 

Since there is a strong likelihood of the application deployment YAML being persisted on the device, it might make sense to keep it since those documents might be used independently of any API calls to get them.

matlec:
> > Thanks @matlec @phil-abb for review I will correct #.1 #.2 #.3 Regarding the rationale of the SUP we can correct it as below. What you people say? Although `ApplicationDeployment` is served over a versioned route (`GET /api/v1/clients/{clientId}/deployments/{deploymentId}/{digest}`), `apiVersion` is retained because: it has lifecycle outside the API scope of the API payloads and may reside either in the OCI registry or in the WFM object store, hence it needs an explicit version
> 
> Fair point. Clearly, I don't have a strong opinion on this because I keep going back and forth 🧇 😄
> 
> Since there is a strong likelihood of the application deployment YAML being persisted on the device, it might make sense to keep it since those documents might be used independently of any API calls to get them.

Haha, the waffle is real 🧇 😄

The "persisted on device" argument is true, but if "persisted and usable outside an API call" is the bar for keeping `apiVersion`, we should also keep `apiVersion` in `DeviceCapabilitiesManifest` and `DeploymentStatusManifest`, both posted by the device, persisted in the WFM. If we accept persistence as discriminator, honestly, I would discard the whole idea of removing `apiVersion` and `kind` from the payload.
Oh, and btw, the current spec doesn't specify persistence for `ApplicationDeployment`s in an OCI registry. We should get rid of that reference in the SUP.

Imho, what sets `ApplicationDescription` genuinely apart is that it has a lifecycle phase with no versioned envelope at all. It's hand-authored, edited and validated as a standalone file before it's ever packaged. In that phase there is nothing to carry the version. Once it's in the OCI registry it's actually redundant there too (this is what I just realized now): the version in the [`mediaType`](https://docs.margo.org/specification/applications/application-registry#margo-specific-media-types) already identifies the contract (🧇). So only the authoring phase would justify keeping it. `ApplicationDeployment` and others are machine-generated and only ever travel over a versioned channel, so sender and receiver can always derive the contract version and persist it outside the structure if needed.

To summarize the options from my point of view:
a) if the discriminator is being able to identify the contract from a structure at rest (detached from the channel), I would keep `apiVersion` and `kind` in all structures
b) if the discriminator is whether a structure is ever handled outside a versioned envelope (no URL and no `mediaType` available), I would keep `apiVersion` and `kind` only for `ApplicationDescription`
c) same as b), but if we want to be future-proof and hedge against a structure later being exchanged over a channel with no version envelope, I would also keep `apiVersion` and `kind` in all structures

My take: I lean towards b). I'd accept c) if we decide that uniformity across all structures is a goal in itself - or if we commit to delivery channels without a versioned envelope (e.g. air-gapped devices receiving deployments offline).
Funny enough, in the original issue I only asked to clarify the accepted values of `apiVersion` 😄

nilanjan-samajdar:
@matlec 

>`ApplicationDeployment` and others are machine-generated and only ever travel over a versioned channel, so sender and receiver can always derive the contract version and persist it outside the structure if needed.

The ApplicationDeployment is not entirely machine-generated, it is customizable to target specific devices identified by `deviceId`. This customization may involve taking inputs from the WFM user as well (not specified).

Additionally, the way the management interface is defined, both objects are "packaged" in the API as yaml, not made part of the OpenAPI body, which hints at their "lifetime" beyond the API.  Hence these 2 need to carry the `apiVersion`.

The `DeviceCaabilities` are not defined as a separate document, they are very much defined as part of the API, hence their versioning follows the associations madefor other APIs as well.

nilanjan-samajdar (approved):

proposals/sup_apiversion_app_description.md
```diff
@@ -0,0 +1,158 @@
...
+
+```yaml
+ApplicationDescription:
+  type: object
+  required: [apiVersion]
+  properties:
+    apiVersion:
+      type: string
+      description: >
+        version of the ApplicationDescription document.
```
phil-abb: I think we need to use "contract" here instead of "document" because "document" could be misinterpreted as literally meaning the version of the specific YAML file and not the ApplicationDescription contract that the YAML file is associated with.

```suggestion
        version of the ApplicationDescription contract.
```

proposals/sup_apiversion_app_description.md
```diff
@@ -0,0 +1,158 @@
...
+      enum:
+        - v1
+      example: v1   
+    # ... other properties
+```
+
+---
+
+
+### Version Progression
```
phil-abb, line 78: I think it makes sense for this section to indicate that these rules apply to both the "apiVersion" in the ApplicationDescription and the `/api/<version>/` in the routes. It's the same rules that apply to both, and I don't think it makes sense to require two separate SUPs to apply the same rules to these two contexts.

phil-abb:
@margo/technical-wg 

The SUP for clarifying handling of `apiVersion` and `kind` is in final review. Please review the proposal and add any questions/comments to the PR.

proposals/sup_apiversion_app_description.md
phil-abb: If you don't want to do this, let me know, and I'll create a companion SUP that can be voted on at the same time for applying these same rules to the API route version.

vireshnavalli: @phil-abb corrected the statement

vireshnavalli:
Hi @phil-abb @ajcraig  could you please schedule it for voting.

ajcraig (approved):
Ready for vote.

phil-abb:
This SUP was approved, so I'm force-merging it in: https://github.com/margo/specification-enhancements/issues/86#issuecomment-5319892130

PR #75, opened 2026-07-05
[DRAFT] Specification Update Proposal: Real-Time

pperycz:
This Specification Update Proposal focuses on following aspects:
* Extending **Applications API** to
  * Specify _workload type_, allowing to hint the `real-time` workload scheduling needs
  * _Improve granularity_ by moving `RequiredResources` under `Component` instead of under `DeploymentProfile`
  * Allow requesting _named groups of resources_ that workload can refer to at runtime
  * Enable requesting specific hybrid _CPU class_
  * Enable requesting isolated/shared _CPU type_
  * Enable requesting shared/exclusive portion of _CPU Cache_
  * Enable requesting _memory bandwidth_ allocation
* Extending **Device Capabilities API** to
  * Expose _workload scheduling_ capabilities of the device (e.g.: can it host real-time workloads)
  * Expose available _CPU kinds_, including hybrid class, isolation type and frequency range
  * Expose _CPU Cache layout_ (topology) with available allocation types (shared/exclusive)
  * Expose available _Memory Allocation types_

> **Note:** This SUP focuses on essential areas needed to enable workload placement of real-time applications: OS-level support for real-time scheduling, and hardware resources isolation enable low-jitter workload execution. To address more complex scenarios additional extensions to Margo Specification will surely be needed.


phil-abb:
[@margo/technical-wg](https://github.com/orgs/margo/teams/technical-wg) A new SUP has been started by @pperycz. If anyone is interested in helping work on fleshing out the details of this SUP, please contact @pperycz

phil-abb:
There are a couple of in-progress SUPs that impact this one:

- [Alternative SUP for device-specific runtimes](https://github.com/margo/specification-enhancements/pull/59) - This SUP proposes changes that put the original "requiredResources" under the new "deviceConstraints" property. This would conflict with the idea of moving the required resources under each component. It's possible that keeping both could make sense. For most common cases, the total required resources are under deviceConstraints.capacityRequirements to make it easier, but if someone needs to be more fine-grained (e.g., real-time workloads), they can also define the specifics under each component. It might be good if this SUP showed an example of what the application description looks like for the non-real-time workload case, so it's clear what is optional and what is required with this proposal. 

- [Resource Discovery](https://github.com/margo/specification-enhancements/pull/66) - @singhmj-1 is still working through some ideas, but this SUP deals with allocatable resources. It covers how the app description requests them and how the device reports what's available and what is consumed. His SUP doesn't deal specifically with these types of resources (memory, CPU, cache), but I think there are still a lot of parallels between them. This SUP doesn't address the tracking of allocated resources. I think we need to make this part very clear. I get this impression that this SUP is proposing that it's the responsibility of the WFM to track allocated resources, but it might make more sense if the device were responsible for this. Since the device is the only one that knows the truth, it seems the device should report which of the allocatable resources is allocated to the WFM instead of the WFM trying to figure it out. @singhmj-1 just published [this SUP](https://github.com/margo/specification-enhancements/pull/76) as well that is broken out from the first one

phil-abb:
The way I see things right now with Margo is that the application description is used to provide information about what is needed and what to deploy, and the device capabilities are used to provide information about what devices can do. Neither the application description nor the device capabilities is used to impact functionality. For example, there are currently no expectations that anything would manipulate my Helm chart in order to be able to deploy it onto a Margo-conformant Kubernetes device apart from passing in the customer-provided parameter values and OTEL collector environment variables. 

Based on this, I would expect that if I wanted to deploy real-time workloads using Helm, my Helm chart would already need to be configured to do this regardless of whether I was deploying to a Margo-conformant device or my own non-Margo Kubernetes cluster. I would not expect Margo to be able to take my helm chart that is not already implemented in a way to be deployable as real-time workloads and manipulate the manifest to make it possible.

The way I see this SUP is enhancements so apps can inform what they are going to need and devices can inform what they can do to support real-time, without introducing functionality that would need to manipulate manifests to make this work. Is my interpretation of this SUP correct? Or would you expect that in order for devices to implement this functionality, they would have to manipulate application manifests before deployment?

A follow-up to that question is, what are the expectations for both the application and the device? 

I do not have experience deploying real-time workloads, but based on my understanding, it's not possible to implement a Helm chart with real-time capabilities to this level without relying on special plugins to make it work. For example, the [NRI Balloons Policy plugin](https://containers.github.io/nri-plugins/stable/docs/resource-policy/policy/balloons.html) you referenced. Would the expectation be that I implement my Helm chart using one of these plugins and need to indicate that I expect that specific plugin to be on the device? Or do you have other expectations on how this would be actualized?

proposals/margo-real-time.md
```diff
@@ -0,0 +1,553 @@
...
+      - `unit` of type `string` with allowed values:
+
+        - `percent` - capping memory usage to `%` value (e.g.: in increments of 10)
+        - `MBps` - capping memory at a specific throughput value
+
+      - `value` of type `integer` holding the actual value of the bandwidth.
+
+    **NOTE** Extending `requiredResources.memory` is a breaking change since it changes from `string` to `object`.
+
+### Extending Device Capabilities API (Topology)
```
singhmj-1, line 317: Interestingly, I have a SUP that also defines a way to extend device capabilities. If you may take a look: https://github.com/margo/specification-enhancements/pull/76 .. and then we can collaborate on this.

chrisgclayton:
I liked this SUP but would propose we look at leveraging the SUP indicated above (https://github.com/margo/specification-enhancements/pull/76) and use that base concept to derive these. 

- I agree with the need to define and target different scheduling (e.g., real-time) and think that the platform schedulers make sense to be a 1st class concept in Margo as it has unique requirements and concepts
- I was wondering if it was just an example or when the named CPU groups were specified, was that to be the name on the device or a name being provided for filling in the variables for the deployment. I believe I know what was intended but would like to confirm here.

Thanks

PR #76, opened 2026-07-07
[DRAFT] bisects DeviceCapabilityManifest into different documents

singhmj-1:
This Specification Update Proposal (SUP) addresses the data coupling and scalability issues identified in [issue](https://github.com/margo/specification/issues/191).

It refactors the DeviceCapabilitiesManifest schema to decouple static device identification from dynamic hardware and runtime capability states, establishing a clean baseline for future capability frameworks.

### DIFF
--- a/proposals/sup_bisect_device_manifest.md
+++ b/proposals/sup_bisect_device_manifest.md
@@ -54,13 +54,15 @@ Identity data rarel changes. Resource related data changes often. Sending them t
 * No independent updates: You cannot version or evolve identity fields and resource fields separately.
 
 ### Problem 2: Limited descriptions for hardware and networks
-The current setup uses basic labels that cannot share real-world details.
+The current setup uses basic labels that cannot share real-world details, as of now:
 
-* A graphics card (GPU) can only list its type and manufacturer. It cannot show its VRAM size, system path, or architecture.
-* A network connection (like CAN Bus) can only list its type. It cannot show its speed, protocol, or channel ID.
+* A graphics card (GPU) lists only type and manufacturer. It doesn't show its VRAM size, system path, or architecture etc.
+* A network connection (like CAN Bus) has only a type field. It doesn't show its speed, protocol, or channel ID etc.
+
+The point to make here is that there would be requirement for a much complexer data structure for different resources and capabilities provided by platforms. It would be better to detach give them their dedicated documents.
 
 ### Problem 3: Basic computing fields cannot grow
-Standard fields like cpu, memory, and storage are treated as rigid, flat text blocks. You cannot expand their rules. They need to become standalone profile files with their own flexible rulebooks.
+Standard fields like cpu, memory, and storage are treated as rigid, flat text blocks. You cannot expand their structure easily. They need to become standalone profile files with their own flexible rulebooks.
 
 ### Problem 4: Hardware types are hardcoded
 Adding a new piece of hardware (like an AI accelerator or secure chip) forces a complete update to the core system specification. This creates a bottleneck. Device makers cannot launch new features until the main community updates the global rules.
--- a/proposals/sup_bisect_device_manifest.md
+++ b/proposals/sup_bisect_device_manifest.md
@@ -3,6 +3,7 @@
 **Status:** Draft  
 **Owner:** @singhmj-1  
 **Depends on:** `sup_device_roles_to_capabilities`
+
 **Note:** This SUP has been extracted from [Resource Discovery and Conflict Resolution](https://github.com/margo/specification-enhancements/pull/66) SUP, as that one has become too complex.
 
 ---
@@ -12,15 +13,16 @@
 The `sup_device_roles_to_capabilities` SUP removed `roles` and promoted all device fields to the `properties` root. The resulting `DeviceCapabilitiesManifest` still conflates two distinct concerns:
 
 - **Device identity** — `id`, `vendor`, `modelNumber`, `serialNumber`. Static after onboarding. Never changes.
-- **Device capabilities** — `cpu`, `memory`, `storage`, `otelCollector`, `supportedRuntimes`, `supportedDeploymentTypes`, `peripherals`, `interfaces`. Dynamic — updated as hardware and runtime state changes.
+- **Device profile information** — `cpu`, `memory`, `storage`, `otelCollector`, `supportedRuntimes`, `supportedDeploymentTypes`, `peripherals`, `interfaces`. Changes primarily when hardware is added/removed or runtime support is modified.
 
 This proposal:
 
-1. Bisects the document into two separate documents with separate endpoints: **DeviceManifest** (identity only) and individual **CapabilityState** documents (one per capability).
-2. Expands the definition of a capability.
+1. Bisects the document into two separate documents with separate endpoints: **DeviceManifest** (identity only) and individual **ProfileState** documents (one per device profile attribute).
+2. Introduces the concept of **Device Profile** — a structured description of all hardware resources, interfaces, peripherals, and services available on a device.
 3. Removes the flat `peripherals` and `interfaces` enumerations entirely.
-4. Promotes all capability data — including base compute fields — into structured `CapabilityState` documents, each posted individually to the capabilities endpoint.
-5. Introduces the `CapabilityState` model inline, as the Capability Definition Framework SUP has not yet been formally introduced.
+4. Promotes all profile data — including base compute fields and services — into structured `ProfileState` documents, each posted individually to the profile endpoint.
+5. Simplifies the `ProfileState` schema to remove unnecessary wrapper fields.
+6. Provides formal schema specifications for each standard profile URI.
 
 ---
 
@@ -53,12 +55,12 @@ This proposal modifies the schema produced by `sup_device_roles_to_capabilities`
 
 ## Reason for Proposal
 
-### Problem 1 — Identity and capabilities are coupled in a single document
+### Problem 1 — Identity and profile information are coupled in a single document
 
-Identity fields (`id`, `vendor`, `modelNumber`, `serialNumber`) never change after onboarding. Capability fields change frequently — whenever workloads are deployed, hardware is added, or runtime configuration changes. Sending both together means:
+Identity fields (`id`, `vendor`, `modelNumber`, `serialNumber`) never change after onboarding. Profile fields change infrequently — only when hardware is added/removed or when runtime support is modified. Sending both together means:
 
-- A capability update forces re-sending identity data that has not changed
-- The WFM cannot distinguish a capability update from an identity change
+- A profile update forces re-sending identity data that has not changed
+- The WFM cannot distinguish an identity change from a profile change
 - The two concerns cannot evolve or be versioned independently
 
 ### Problem 2 — `peripherals` and `interfaces` use closed enumerations that cannot express structured state
@@ -77,38 +79,44 @@ Cannot express VRAM, device path, allocation state, or compute architecture. A C
 
 Cannot express baud rate, protocol, channel identity, or allocation state.
 
-### Problem 3 — Base compute fields are not capabilities in the structured sense
+### Problem 3 — Base compute fields lack a structured model for extension
 
-Fields like `cpu`, `memory`, `storage`, `otelCollector`, `supportedRuntimes`, and `supportedDeploymentTypes` are currently special-cased flat/scalar fields. There is no mechanism to extend their schema, express allocation state, or update them independently of each other. Treating them as first-class `CapabilityState` entries resolves this.
+Fields like `cpu`, `memory`, `storage`, `otelCollector`, `supportedRuntimes`, and `supportedDeploymentTypes` are currently special-cased flat/scalar fields. There is no mechanism to extend their schema, express allocation metadata, or document their update semantics. Treating them as first-class `ProfileState` entries with formal schema specifications resolves this.
 
 ---
 
-## Expanding the Definition of Capability
+## Introducing Device Profile
 
-So far, the definition of a capability is restrictive, but this framework expands it:
+This proposal introduces the concept of a **Device Profile** — a comprehensive description of all hardware resources, communication interfaces, physical peripherals, and platform services available on a device.
 
-> A capability represents wide variety of **platform-managed concerns** 
-including the following:
-> * "Hardware resources" such as GPUs, cameras, fieldbus channels etc.
-> * "Software capabilities" such as persistent storage, pre-installed 
-runtimes etc.
-> * "Platform services" such as api gateways, security vaults, otel collectors 
-endpoints, pki, message buses etc.
+A **Device Profile** encompasses:
 
+- **Resources** (CPU, Memory, Storage) — Quantifiable hardware resources
+- **Interfaces** (Ethernet, WiFi, CAN Bus, USB) — Communication and connectivity mechanisms
+- **Peripherals** (GPU, Camera, Microphone) — Specialized hardware components
+- **Services** (Secret Vault, DNS Server, OTeL Collector) — Platform-managed services
+- **Capabilities** (Video encoding, AI acceleration, observability collection) — Functional abilities enabled by hardware, services, or software
 
-### Capability Model
-The current spec treats a "capability" as a flat enumeration value — a GPU is `{ "type": "gpu" }`, an interface is `{ "type": "ethernet" }`. This model cannot express runtime state, allocation, or structured metadata.
+### Key Insight: Things vs. Capabilities
 
-This proposal introduces a richer capability model. A **CapabilityState** is a structured document that:
+A crucial distinction: not all profile attributes are themselves *capabilities*. Rather, they are platform-managed *resources* and *services* that collectively form the device's profile and enable certain capabilities:
 
-- References a **capability URI** — a globally unique identifier for the capability type (e.g. `capability.margo.org/hardware/gpu`)
-- Carries a **spec** — a capability-specific object whose schema is defined by the capability type
+- A **GPU** (peripheral) is not a capability, but it *enables* video encoding and AI acceleration (capabilities)
+- A **Secret Vault** (service) is not a capability, but it *enables* secure credential management (capability)
+- A **DNS Server** (service) is not a capability, but it *enables* IP address assignment (capability)
+- An **Ethernet interface** (interface) is not a capability, but it *enables* network communication (capability)
 
-The capability URI is the key. It identifies not just the type of capability, but the schema that governs the `spec` content. This allows:
+By treating all of these under the unified **Device Profile** concept, we can describe the complete environment available to workloads without conflating the *providers* of capabilities with the *capabilities* themselves.
 
-- Vendors to define their own capability types under their own URI namespace (e.g. `capability.acme.com/fieldbus/profinet`)
-- The WFM to validate `spec` content against the schema for the referenced URI
-- Capabilities to be updated, added, or removed independently — one `CapabilityState` per capability, one request per update
+### Pattern Extensibility
+
+This Device Profile pattern is designed to be reusable beyond devices. Similar profiles could be defined for:
+
+- **Workload Profile** — describing workload requirements and constraints
+- **Workflow Profile** — describing workflow execution environment needs
+- **Gateway Profile** — describing gateway-specific resources and services
+
+This creates a consistent mental model across the Margo ecosystem.
 
 ---
 
@@ -124,15 +132,25 @@ POST   /api/v1/clients/{clientId}/devices/{deviceId}
 PUT    /api/v1/clients/{clientId}/devices/{deviceId}
 DELETE /api/v1/clients/{clientId}/devices/{deviceId}
 
-# Individual capability states
-POST   /api/v1/clients/{clientId}/devices/{deviceId}/capabilities
-PUT    /api/v1/clients/{clientId}/devices/{deviceId}/capabilities
-DELETE /api/v1/clients/{clientId}/devices/{deviceId}/capabilities
+# Device profile state
+POST   /api/v1/clients/{clientId}/devices/{deviceId}/profile
+PUT    /api/v1/clients/{clientId}/devices/{deviceId}/profile
+DELETE /api/v1/clients/{clientId}/devices/{deviceId}/profile
 ```
 
-**Ordering requirement:** `DeviceManifest` MUST be registered before any `CapabilityState` can be submitted for the same `deviceId`. If the WFM receives a `CapabilityState` for an unknown `deviceId`, it MUST reject with `404 Not Found`.
+**Ordering requirement:** `DeviceManifest` MUST be registered before any `ProfileState` can be submitted for the same `deviceId`. If the WFM receives a `ProfileState` for an unknown `deviceId`, it MUST reject with `404 Not Found`.
 
-**Individual updates:** Each POST or PUT to the capabilities endpoint carries exactly one `CapabilityState` document. The WFM uses `metadata.id` to identify which capability is being created or updated. Multiple capabilities are registered by sending multiple requests.
+**Update granularity:**
+- **Individual updates (default):** Each POST or PUT to the profile endpoint carries exactly one `ProfileState` document. The WFM uses `id` to identify which profile attribute is being created or updated. Multiple attributes are registered by sending multiple requests.
+- **Bulk updates (optional):** Alternatively, clients MAY submit an array of `ProfileState` documents in a single request to reduce chattiness during initial onboarding:
+  ```json
+  [
+    { "id": "deviceprofile.margo.org/resource/cpu", "spec": { ... } },
+    { "id": "deviceprofile.margo.org/resource/memory", "spec": { ... } },
+    { "id": "deviceprofile.margo.org/peripherals/gpu", "spec": { ... } }
+  ]
+  ```
+  The WFM processes all items atomically: either all succeed or all fail.
 
 **Gateway hierarchy:** The `{deviceId}` path hierarchy encoding (e.g. `gateway1/deviceA`) is retained unchanged on both endpoint groups.
 
@@ -148,7 +166,7 @@ DELETE /api/v1/clients/{clientId}/devices/{deviceId}/capabilities
 | `400 Bad Request` | Missing or invalid `content-digest` header, or malformed request body. |
 | `401 Unauthorized` | Signature verification failed. |
 | `403 Forbidden` | Client certificate is not trusted or has been revoked. |
-| `404 Not Found` | No client with the given `clientId` found, or no `DeviceManifest` registered for the given `deviceId` (capabilities endpoint only). |
+| `404 Not Found` | No client with the given `clientId` found, or no `DeviceManifest` registered for the given `deviceId` (profile endpoint only). |
 | `422 Unprocessable Content` | Request body includes a semantic error. |
 
 ---
@@ -163,14 +181,10 @@ Carries device identity only. Sent once at onboarding. Updated only if identity 
 
 ```json
 {
-  "apiVersion": "margo.org/v1",
-  "kind": "DeviceManifest",
-  "properties": {
-    "id": "northstarida.xtapro.k8s.edge",
-    "vendor": "Northstar Industrial Devices",
-    "modelNumber": "332ANZE1-N1",
-    "serialNumber": "PF45343-AA"
-  }
+  "id": "northstarida.xtapro.k8s.edge",
+  "vendor": "Northstar Industrial Devices",
+  "modelNumber": "332ANZE1-N1",
+  "serialNumber": "PF45343-AA"
 }
 ```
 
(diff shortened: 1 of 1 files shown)


phil-abb:
@singhmj-1 I'm a bit unclear what the intention is between this SUP and [the other one you started](https://github.com/margo/specification-enhancements/pull/66). This SUP seems to cover a lot of the same concepts that were started in other SUP. What's the plan for the other SUP?

singhmj-1:
@phil-abb The other SUP evolved to a point where it was carrying too many concerns simultaneously. The plan is to use this new SUP to establish the clean DeviceManifest / CapabilityState split first. I will then update the other SUP to depend on this one, and removing any redundant content from there as well.

phil-abb:
[@margo/technical-wg](https://github.com/orgs/margo/teams/technical-wg) A new SUP has been started by @singhmj-1 . If anyone is interested in helping work on fleshing out the details of this SUP, please contact @singhmj-1 

This SUP has been broken out from the [previous SUP](https://github.com/margo/specification-enhancements/pull/66) @singhmj-1 had started and would be a precursor to that one. The other SUP will be refined following this one.

proposals/sup_bisect_device_manifest.md
```diff
@@ -0,0 +1,457 @@
...
+}
+```
+
+---
+
+## Reason for Proposal
+
+### Problem 1 — Identity and capabilities are coupled in a single document
+
+Identity fields (`id`, `vendor`, `modelNumber`, `serialNumber`) never change after onboarding. Capability fields change frequently — whenever workloads are deployed, hardware is added, or runtime configuration changes. Sending both together means:
```
phil-abb: >Capability fields change frequently

Based on the current specification, I would say this is overstated. While I agree that some things might change occasionally (especially if it's a virtual device), it's not like these things will be changing weekly or probably even monthly.

Maybe this statement is projecting your changes on top of it. So, if you intended to use the device capabilities to report current allocations, then this would require it to be sent more frequently, but that's because of a change being proposed and not because of how it's currently defined.

proposals/sup_bisect_device_manifest.md
```diff
@@ -0,0 +1,457 @@
...
+
```
phil-abb, line 55: I understand the point that it doesn't make sense to keep sending information that is static every time something unrelated changes, but there's much information being sent with these four fields (granted, more might be added in the future). 

I agree this information is not related to capabilities, but that doesn't mean seperate documents would need to be created. We could just as easily change the name and structure of this document to indicate it is a collection of device information and not only "capabilities," as the current name suggests.

proposals/sup_bisect_device_manifest.md
```diff
@@ -0,0 +1,457 @@
...
+Each capability is submitted as an individual `CapabilityState` document. There is no wrapping manifest — the `CapabilityState` is the document.
+
+**Endpoint:** `POST /api/v1/clients/{clientId}/devices/{deviceId}/capabilities`
+
+The WFM uses `metadata.id` to identify which capability is being registered or updated. Submitting a `CapabilityState` for a `capability` URI that already exists for the device replaces the previous state.
+
+**Document structure:**
+
+```json
```
phil-abb, line 319: Based on the other discussions we've been having about "apiVersion and "kind", do we need these here? Also, I don't know that including the "metadata" section is needed. Could it just be:

```json
{
  "id": "<capability-uri>"
  "spec": {
    // capability-specific state object
  }
}
```

proposals/sup_bisect_device_manifest.md
```diff
@@ -0,0 +1,457 @@
...
+
+### CapabilityState
+
+Each capability is submitted as an individual `CapabilityState` document. There is no wrapping manifest — the `CapabilityState` is the document.
+
+**Endpoint:** `POST /api/v1/clients/{clientId}/devices/{deviceId}/capabilities`
+
+The WFM uses `metadata.id` to identify which capability is being registered or updated. Submitting a `CapabilityState` for a `capability` URI that already exists for the device replaces the previous state.
+
+**Document structure:**
```
phil-abb: Is there a reason not to make this an array instead of requiring a bunch of individual API calls? Based on the proposal, this will result in a lot of chattiness, primarily during the initial onboarding, where all these capabilities are sent to the WFM.  Would it make sense instead for this to be an array so all the capabilities that need to be sent could be sent at the same time?

proposals/sup_bisect_device_manifest.md
```diff
@@ -0,0 +1,457 @@
...
+
+### DeviceManifest
+
+Carries device identity only. Sent once at onboarding. Updated only if identity information changes.
+
+**Endpoint:** `POST /api/v1/clients/{clientId}/devices/{deviceId}`
+
+**Example:**
+
+```json
```
phil-abb, line 291: Same thing, based on the discussions over "kind" and "apiVersion", can't we remove them from the payload?

Also, is there a reason to have the "properties" subgrouping? Is there something other than "properties" that is planned to be added?

proposals/sup_bisect_device_manifest.md
```diff
@@ -0,0 +1,457 @@
...
+
+---
+
+### Capability URI Namespace
+
+Capability URIs follow a reverse-DNS convention:
+
+| Namespace | Owner | Example |
+|-----------|-------|---------|
+| `capability.margo.org/*` | Margo specification — standardised capabilities | `capability.margo.org/hardware/gpu` |
```
phil-abb: A repeat of a previous comment (see https://github.com/margo/specification-enhancements/pull/66#issuecomment-4632472032), but a GPU is not a capability. GPUs enable certain capabilities, but it's not a capability. 

GPU is just an example, but I think we have been mixing the terms from the beginning. We're calling the "thing" the capability instead of the functionality that the "thing" allows for.

proposals/sup_bisect_device_manifest.md
```diff
@@ -0,0 +1,457 @@
...
+| `capability.<vendor-domain>/*` | Vendor-defined capabilities | `capability.acme.com/fieldbus/profinet` |
+
+The WFM MUST accept any well-formed URI in `metadata.id`. It MUST validate `spec` content against the schema for the referenced URI if a schema is registered. If no schema is registered for the URI, the WFM MUST store the `spec` as-is without validation.
+
+---
+
+### Standard Capability URIs Introduced by This SUP
+
+The following capability URIs replace the flat fields from the baseline schema. All are under the `capability.margo.org` namespace.
```
phil-abb: I believe this SUP is proposing that the following become official Margo definitions. If this is the case, we should probably define the parameters for each of them. The data types, required/optional, acceptable values, description, etc., as part of this SUP

proposals/sup_bisect_device_manifest.md
```diff
@@ -0,0 +1,457 @@
...
+    "id": "capability.acme.com/fieldbus/canbus"
+  },
+  "spec": {
+    "channels": [
+      {
+        "channelId": "can0",
+        "baudRate": 500000,
+        "protocol": "CANopen",
+        "allocated": false,
+        "allocatedBy": ""
```
phil-abb: I commented on the need to define the parameters for these types above, so I'm not sure what the allowed properties would be, but I think the "allocated" and "allocatedBy" properties duplicate information. I don't think the "allocated" is needed because this will always be false when "allocatedBy" is empty (or not sent), and always true when "allocatedBy" is set.

phil-abb:
I think there are some interesting ideas here. I agree our deviceCapabilities document is currently a bit of a mish-mash of things and not really named correctly for what it provides. 

@pperycz's [real-time workload SUP](https://github.com/margo/specification-enhancements/pull/75) is adding even more requirements for being able to provide fine-grained details about resources, so I think there could be some opportunities for collaboration here.

proposals/sup_bisect_device_manifest.md
phil-abb: This doesn't mean we cannot use the "thing" as part of the ID, but in the cases we do, we shouldn't call it a capability. We don't have to keep using "capability" just because we used it originally if there is a better term to use.

proposals/sup_bisect_device_manifest.md
singhmj-1: @phil-abb  I see what you mean now. If cpu and memory only mean "total hardware capacity," then yes, they won't change often.

But from a practical standpoint, just knowing the total capacity doesn't help much unless we talk about what resources are free or occupied. I was operating under the assumption that resource fields should reflect real-time or frequent updates whenever the device sees a state change.

Since this requirement will vary heavily from vendor to vendor, decoupling the documents gives us the flexibility to support those dynamic allocation reports use case as well.

phil-abb: No, that's the point of having the [observability stuff](https://docs.margo.org/specification/observability/collecting-workload-observability-data). It reports the current usage for these [typical resources](https://docs.margo.org/specification/observability/collecting-workload-observability-data#workload-observability-default-telemetry)

proposals/sup_bisect_device_manifest.md
singhmj-1: I agree that they could technically live under the same document.

However, to me, the completely separate documents provide a much better separation of concerns and are cleaner to understand than putting everything under a single object. It will prevent the two scopes from bleeding into each other as the spec grows.

proposals/sup_bisect_device_manifest.md
singhmj-1: I'll update it tomorrow. Thanks!

proposals/sup_bisect_device_manifest.md
singhmj-1: Yeah, this will be helpful. Will update the proposal. Thanks!

proposals/sup_bisect_device_manifest.md
singhmj-1: Actually, this proposal expands the definition of a `Capability`, [check this](https://github.com/margo/specification-enhancements/pull/76/files#diff-5c5601aef2849436e9b6e2d61a1032fac4799d014852a7e501fc70f426ea6789R86) . 
and as per it a GPU is a platform-concern, and is considered as a hardware capability.

Btw, I actually heavily considered introducing a broader term like `PlatformConcern` or `PlatformFeature` or `Resource` to address wider range of these things available on a system. But I chose to extend the existing `Capability` terminology to keep the conceptual surface area small for us.

proposals/sup_bisect_device_manifest.md
singhmj-1: But if the WFM needs to make an orchestration decision based on whether a GPU is occupied on a device, will observability data really help there?

Observability metrics show real-time resource utilization, but they don't show logical assignments. If a workload is currently idling, metrics might show 0% usage, but the resource is still technically claimed. I am coming from the standpoint that WFMs should be enabled to see these allocation states to make accurate scheduling decisions, **if they want to**.

singhmj-1: I think I missed a key point on how this works under the hood: `CapabilityState` tracks the logical assignment of a resource, along-with metrics.

phil-abb: No, the observability data will not show everything. The list of what metrics are expected to be there by default is listed at the bottom of the page: https://docs.margo.org/specification/observability/collecting-workload-observability-data#workload-observability-default-telemetry

There is a difference between indicating if something allocatable is allocated versus showing usage metrics. I agree it probably makes sense to use this API to inform on what is allocated and allocatable, but I don't think it makes sense to try to use it to communicate those real-time usage metrics.

proposals/sup_bisect_device_manifest.md
phil-abb: I think I would prefer an overarching term such as DeviceProfile instead of DeviceCapabilities, and then the DeviceProfile can have things like Resource (CPU, Memory, Storage), Interfaces (Ethernet, WiFi, USB), Peripherals (GPU, Camera, Microphone), and Capabilities (Video encoding, AI acceleration). 

For example, the endpoint would be `/api/v1/clients/{clientId}/devices/{deviceId}/profile` and IDs could be:
- deviceprofile.margo.org/resource/cpu
- deviceprofile.margo.org/resource/memory
- deviceprofile.margo.org/resource/storage
- deviceprofile.margo.org/capability/observability
- deviceprofile.margo.org/capability/runtimes
- deviceprofile.margo.org/peripherals/gpu
- deviceprofile.margo.org/interface/ethernet
- deviceprofile.margo.org/interface/canbus

proposals/sup_bisect_device_manifest.md
singhmj-1: I have updated the SUP content.

proposals/sup_bisect_device_manifest.md
singhmj-1: I liked the idea of a device profile. I have updated the SUP content and introduced the concept of a DeviceProfile, can you please take some time to review again? Thanks!

phil-abb: @singhmj-1 - I should have some time on Friday to review the latest changes 👍

proposals/sup_bisect_device_manifest.md
```diff
@@ -0,0 +1,819 @@
...
+```
+
+This proposal completely eliminates this combined format and rebuilds the way features are reported from scratch.
+
+---
+
+## Reason for Proposal
+
+### Problem 1: Identity and resources are mixed together
+Identity data rarel changes. Resource related data changes often. Sending them together creates major issues:
```
phil-abb, line 50:
```suggestion
Identity data rarely changes. Resource related data changes often. Sending them together creates major issues: 
```

proposals/sup_bisect_device_manifest.md
```diff
@@ -0,0 +1,819 @@
...
+* Confusion: The system cannot separate a serial number fix from a hardware change.
+* No independent updates: You cannot version or evolve identity fields and resource fields separately.
+
+### Problem 2: Limited descriptions for hardware and networks
+The current setup uses basic labels that cannot share real-world details, as of now:
+
+* A graphics card (GPU) lists only type and manufacturer. It doesn't show its VRAM size, system path, or architecture etc.
+* A network connection (like CAN Bus) has only a type field. It doesn't show its speed, protocol, or channel ID etc.
+
+The point to make here is that there would be requirement for a much complexer data structure for different resources and capabilities provided by platforms. It would be better to detach give them their dedicated documents.
```
phil-abb, line 62:
```suggestion
The point to make here is that there would be requirements for a much more complex data structure for the different resources and capabilities provided by platforms. It would be better to detach them to give each their dedicated documents.
```

proposals/sup_bisect_device_manifest.md
```diff
@@ -0,0 +1,819 @@
...
+    "platformState": { }
+  },
+  {
+    "profileDefinitionId": "deviceprofile.margo.org/peripherals/gpu",
+    "platformState": { }
+  }
+]
+```
+
+The WFM processes all items atomically: either all succeed or all fail. When it fails, the WFM MUST return a `422 Unprocessable Content` response identifying which `profileDefinitionId` caused the failure:
```
phil-abb, line 248: We probably want to think about how to incorporate the proposed changes to use RFC 9457: https://github.com/margo/specification-enhancements/pull/73

 There's a good chance this SUP will get approved. We can probably wait until the voting, but it might be good to start thinking about it.

proposals/sup_bisect_device_manifest.md
```diff
@@ -0,0 +1,819 @@
...
+  "profileDefinitionId": "deviceprofile.margo.org/service/otel-collector",
+  "description": "Profile Definition for the OpenTelemetry collector service",
+  "scope": "device",
+  "category": "service",
+  "schemaVersion": "1.0.0",
+  "platformStateSchema": {
+    "type": "object",
+    "required": ["endpoint"],
+    "properties": {
+      "endpoint": {
```
phil-abb, line 650: I'm not sure what information makes sense to send for the otel collector other than an indication of it being available. There's really nothing the WFM can do with the endpoint, so I don't think it's helpful to send.

phil-abb: Maybe we can start with a simple boolean and think about it afterwards if this gets approved. The only thing I can think of right now that might be useful is client certificate information for enabling mTLS, but I don't know if we'd want to use this for that.

proposals/sup_bisect_device_manifest.md
```diff
@@ -0,0 +1,819 @@
...
+  "category": "capability",
+  "schemaVersion": "1.0.0",
+  "platformStateSchema": {
+    "type": "object",
+    "required": ["containerRuntimes", "deploymentTypes"],
+    "properties": {
+      "containerRuntimes": {
+        "type": "array",
+        "minItems": 1,
+        "items": { "type": "string" },
```
phil-abb, line 688: This should be an enum with just "oci" for now.

proposals/sup_bisect_device_manifest.md
```diff
@@ -0,0 +1,819 @@
...
+        "description": "List of supported container runtime standards (e.g. oci)"
+      },
+      "deploymentTypes": {
+        "type": "array",
+        "minItems": 1,
+        "items": { "type": "string" },
```
phil-abb, line 694: This should be an enum with just "helm" and "compose" for now. Depending on the voting results for the custom runtime support, "custom" might be added.

proposals/sup_bisect_device_manifest.md
```diff
@@ -0,0 +1,819 @@
...
+```
+ProfileDefinition  →  defines the contract (schema, scope, ownership, category)
+ProfileState       →  runtime data conforming to that contract
+```
+
+Devices send out many `ProfileState` documents after onboarding and update them when there are changes. Every `ProfileState` must conform to the rules written in its matching `ProfileDefinition` blueprint.
+
+---
+
+### ProfileDefinition
```
phil-abb, line 85: It might be helpful to come up with a starting list of guiding principles for these profile definitions. 

For example, I think it should be a goal to try to keep the required properties to the absolute minimum needed to enable basic interoperability.

It would be good to make it easy to meet the standard use case, but be flexible enough to allow advanced use cases to be met. I'm thinking of the real-time workload use case here. We don't want to require devices with no intention of supporting real-time workloads to have to provide additional unnecessary details just because we want to support real-time workloads.

proposals/sup_bisect_device_manifest.md
phil-abb: @singhmj-1 - I like the latest updates a lot. I like how this idea is coming together. 

I'd be interested in hearing what other @margo/technical-wg members think of it.

phil-abb:
@singhmj-1 / @pperycz / @rchavez11b  - I've used some AI assistance to take the ideas expressed in this SUP and merge them with the ideas expressed in the [real-time workload SUP](https://github.com/margo/specification-enhancements/pull/75) just to get an idea of how things might look. 

I think this gives a good concept for how this could work in practice. What do you guys think?

## Profile Definitions

**CPU Profile Definition**
```json
{
  "profileDefinitionId": "deviceprofile.margo.org/resource/cpu",
  "description": "CPU resource profile with minimal baseline fields and optional real-time topology hints.",
  "scope": "device",
  "category": "resource",
  "schemaVersion": "1.0.0",
  "platformStateSchema": {
    "type": "object",
    "required": [
      "cpus"
    ],
    "properties": {
      "cpus": {
        "type": "array",
        "minItems": 1,
        "items": {
          "type": "object",
          "required": [
            "cores"
          ],
          "properties": {
            "cores": {
              "type": "number",
              "minimum": 0.001,
              "description": "CPU cores delegated to Margo (supports fractional cores)."
            },
            "architecture": {
              "type": "string",
              "enum": [
                "amd64",
                "arm64",
                "arm",
                "x86",
                "x86_64"
              ],
              "description": "Optional architecture from current DeviceCapabilities model."
            },
            "class": {
              "type": "string",
              "enum": [
                "performance",
                "efficiency",
                "low-power"
              ],
              "description": "Optional real-time SUP extension for hybrid CPU class."
            },
            "type": {
              "type": "string",
              "enum": [
                "shared",
                "isolated"
              ],
              "description": "Optional real-time SUP extension for scheduling isolation intent."
            },
            "name": {
              "type": "string",
              "description": "Optional named CPU group usable by application workloads."
            },
            "frequency": {
              "type": "object",
              "properties": {
                "maxMHz": {
                  "type": "integer",
                  "minimum": 1
                },
                "baseMHz": {
                  "type": "integer",
                  "minimum": 1
                }
              },
              "additionalProperties": false,
              "description": "Optional max/base frequency details from real-time SUP examples."
            },
            "cpuset": {
              "type": "string",
              "description": "Optional CPU set representation (for example, 0-3,8)."
            },
            "numaNode": {
              "type": "integer",
              "minimum": 0,
              "description": "Optional NUMA node association."
            }
          },
          "additionalProperties": true
        }
      }
    },
    "additionalProperties": true
  }
}
```

**CPU Cache Profile Definition**
```json
{
  "profileDefinitionId": "deviceprofile.margo.org/resource/cpu-cache",
  "description": "CPU cache resource profile with minimal cache descriptors and optional allocation capabilities for real-time workloads.",
  "scope": "device",
  "category": "resource",
  "schemaVersion": "1.0.0",
  "platformStateSchema": {
    "type": "object",
    "required": [
      "caches"
    ],
    "properties": {
      "caches": {
        "type": "array",
        "minItems": 1,
        "items": {
          "type": "object",
          "required": [
            "level",
            "size"
          ],
          "properties": {
            "level": {
              "type": "string",
              "enum": [
                "L3",
                "L2",
                "L1d",
                "L1i"
              ],
              "description": "Cache level requested in the real-time SUP."
            },
            "size": {
              "type": "string",
              "pattern": "^[0-9]+(Ki|Mi|Gi|KB|MB|GB)$",
              "description": "Cache size with binary or decimal unit."
            },
            "allocationTypes": {
              "type": "array",
              "items": {
                "type": "string",
                "enum": [
                  "exclusive",
                  "shared"
                ]
              },
              "minItems": 1,
              "uniqueItems": true,
              "description": "Optional real-time SUP extension describing supported cache allocation modes."
            },
            "scope": {
              "type": "string",
              "description": "Optional affinity scope (for example, package, core-group, numa-node)."
            }
          },
          "additionalProperties": true
        }
      }
    },
    "additionalProperties": true
  }
}
```

**Memory Profile Definition**
```json
{
  "profileDefinitionId": "deviceprofile.margo.org/resource/memory",
  "description": "Memory resource profile with baseline size and optional real-time bandwidth capability details.",
  "scope": "device",
  "category": "resource",
  "schemaVersion": "1.0.0",
  "platformStateSchema": {
    "type": "object",
    "required": [
      "size"
    ],
    "properties": {
      "size": {
        "type": "string",
        "pattern": "^[0-9]+(Ki|Mi|Gi|Ti)$",
        "description": "Total memory available to Margo workloads."
      },
      "bandwidthAllocationTypes": {
        "type": "array",
        "items": {
          "type": "string",
          "enum": [
            "percent-throttling",
            "rate-limiting"
          ]
        },
        "minItems": 1,
        "uniqueItems": true,
        "description": "Optional real-time SUP extension for bandwidth control capabilities."
      },
      "bandwidth": {
        "type": "object",
        "properties": {
          "unit": {
            "type": "string",
            "enum": [
              "percent",
              "MBps"
            ]
          },
          "value": {
            "type": "integer",
            "minimum": 1
          }
        },
        "required": [
          "unit",
          "value"
        ],
        "additionalProperties": false,
        "description": "Optional concrete memory bandwidth capacity representation."
      },
      "numaNodes": {
        "type": "array",
        "items": {
          "type": "object",
          "required": [
            "id",
            "size"
          ],
          "properties": {
            "id": {
              "type": "integer",
              "minimum": 0
            },
            "size": {
              "type": "string",
              "pattern": "^[0-9]+(Ki|Mi|Gi|Ti)$"
            },
            "bandwidth": {
              "type": "object",
              "properties": {
                "unit": {
                  "type": "string",
                  "enum": [
                    "percent",
                    "MBps"
                  ]
                },
                "value": {
                  "type": "integer",
                  "minimum": 1
                }
              },
              "required": [
                "unit",
                "value"
              ],
              "additionalProperties": false
            }
          },
          "additionalProperties": true
        },
        "description": "Optional NUMA-level memory topology details."
      }
    },
    "additionalProperties": true
  }
}
```

**Workload Scheduling Profile Definition**
```json
{
  "profileDefinitionId": "deviceprofile.margo.org/capability/workload-scheduling",
  "description": "Workload scheduling capability profile with minimal defaults and optional advanced real-time scheduling metadata.",
  "scope": "device",
  "category": "capability",
  "schemaVersion": "1.0.0",
  "platformStateSchema": {
    "type": "object",
    "required": [
      "schedulingTypes"
    ],
    "properties": {
      "schedulingTypes": {
        "type": "array",
        "items": {
          "type": "string",
          "enum": [
            "background",
            "best-effort",
            "deadline",
            "real-time"
          ]
        },
        "minItems": 1,
        "uniqueItems": true,
        "description": "Supported workload scheduling modes; best-effort-only is valid for non-real-time devices."
      },
      "realTimePriorityRange": {
        "type": "object",
        "properties": {
          "min": {
            "type": "integer",
            "minimum": 1,
            "maximum": 99
          },
          "max": {
            "type": "integer",
            "minimum": 1,
            "maximum": 99
          }
        },
        "required": [
          "min",
          "max"
        ],
        "additionalProperties": false,
        "description": "Optional static priority range for SCHED_FIFO/SCHED_RR workloads."
      },
      "niceRange": {
        "type": "object",
        "properties": {
          "min": {
            "type": "integer",
            "minimum": -20,
            "maximum": 19
          },
          "max": {
            "type": "integer",
            "minimum": -20,
            "maximum": 19
          }
        },
        "required": [
          "min",
          "max"
        ],
        "additionalProperties": false,
        "description": "Optional nice-range information for best-effort/background workloads."
      },
      "deadlineScheduling": {
        "type": "object",
        "properties": {
          "supported": {
            "type": "boolean"
          },
          "units": {
            "type": "string",
            "enum": [
              "microseconds",
              "milliseconds"
            ]
          }
        },
        "required": [
          "supported"
        ],
        "additionalProperties": false,
        "description": "Optional support metadata for deadline scheduling controls."
      }
    },
    "additionalProperties": true
  }
}
```
## Examples

**Device profile for a basic non-real-time capable device**
```json
[
  {
    "profileDefinitionId": "deviceprofile.margo.org/resource/cpu",
    "platformState": {
      "cpus": [
        {
          "cores": 4
        }
      ]
    }
  },
  {
    "profileDefinitionId": "deviceprofile.margo.org/resource/memory",
    "platformState": {
      "size": "16Gi"
    }
  },
  {
    "profileDefinitionId": "deviceprofile.margo.org/capability/workload-scheduling",
    "platformState": {
      "schedulingTypes": [
        "best-effort"
      ]
    }
  }
]
```

**Device profile for a real-time workload capable device**
```json
[
  {
    "profileDefinitionId": "deviceprofile.margo.org/resource/cpu",
    "platformState": {
      "cpus": [
        {
          "name": "rt-performance-isolated",
          "class": "performance",
          "cores": 1,
          "type": "isolated",
          "architecture": "amd64",
          "frequency": {
            "maxMHz": 4700,
            "baseMHz": 1900
          },
          "cpuset": "0",
          "numaNode": 0
        },
        {
          "name": "background-efficiency-shared",
          "class": "efficiency",
          "cores": 8,
          "type": "shared",
          "architecture": "amd64",
          "frequency": {
            "maxMHz": 3700,
            "baseMHz": 1600
          },
          "cpuset": "4-11",
          "numaNode": 0
        }
      ]
    }
  },
  {
    "profileDefinitionId": "deviceprofile.margo.org/resource/cpu-cache",
    "platformState": {
      "caches": [
        {
          "level": "L3",
          "size": "18432KB",
          "allocationTypes": [
            "exclusive",
            "shared"
          ],
          "scope": "package"
        },
        {
          "level": "L2",
          "size": "3072KB",
          "allocationTypes": [
            "exclusive"
          ],
          "scope": "core-group"
        }
      ]
    }
  },
  {
    "profileDefinitionId": "deviceprofile.margo.org/resource/memory",
    "platformState": {
      "size": "16Gi",
      "bandwidthAllocationTypes": [
        "percent-throttling",
        "rate-limiting"
      ],
      "bandwidth": {
        "unit": "MBps",
        "value": 20480
      },
      "numaNodes": [
        {
          "id": 0,
          "size": "8Gi",
          "bandwidth": {
            "unit": "percent",
            "value": 30
          }
        },
        {
          "id": 1,
          "size": "8Gi",
          "bandwidth": {
            "unit": "percent",
            "value": 30
          }
        }
      ]
    }
  },
  {
    "profileDefinitionId": "deviceprofile.margo.org/capability/workload-scheduling",
    "platformState": {
      "schedulingTypes": [
        "best-effort",
        "real-time",
        "deadline",
        "background"
      ],
      "realTimePriorityRange": {
        "min": 1,
        "max": 99
      },
      "niceRange": {
        "min": -20,
        "max": 19
      },
      "deadlineScheduling": {
        "supported": true,
        "units": "milliseconds"
      }
    }
  }
]
```

phil-abb:
@matlec, I'd be interested in getting your take on this SUP if you have some time to look it over.

proposals/sup_bisect_device_manifest.md
nilanjan-samajdar: > It would be good to make it easy to meet the standard use case, but be flexible enough to allow advanced use cases to be met. I'm thinking of the real-time workload use case here. We don't want to require devices with no intention of supporting real-time workloads to have to provide additional unnecessary details just because we want to support real-time workloads.

I agree, it would mean that we have a "default" set of  `ProfileDefinition` already specified in the examples, to start with with scope for the community to improve further.
We may need to set the ground-rules of how this flexibility is exercised - 
- Margo specifies a default set of `ProfileDefinition`
- Vendors may add more to these definitions, but they need to register these `ProfileDefinition` with the WFM, hence giving the vocabulary for  resources and their interpretation.
- The `ProfileDefinition` is used to implement (create and parse) the `ProfileState`, without using any hard-coded knowledge, by the WFM and WFM Client. The vendors (WFM and Device) need to be adopt the flexible schema specified as the `ProfileDefinition`

chrisgclayton:
I like the concept of the with breaking it out. This provides a better level of control and IMO the ability to provide some well known gestures associated with specific categories (e.g., a WFM may treat resources different that peripherals etc.). 

In reading this I noted that the examples were using the query strings that included the {clientId}. I thought we were moving away from this with the MIAF SUP as the client Id is already known from the SVID and if they conflicted there is no required conflict resolution?

PR #77, opened 2026-07-13
Read receipts for desired state

matlec:
This SUP adds read receipts for desired state: one required field on the `DeploymentStatusManifest`, `observedManifestVersion`, where the client reports the `manifestVersion` it has reconciled a deployment against. It lets the WFM tell for each deployment whether the client has picked up the current change. This is a breaking change.

### DIFF
--- a/proposals/sup_desired_state_read_receipts.md
+++ b/proposals/sup_desired_state_read_receipts.md
@@ -115,6 +115,9 @@ Semantics:
 - `observedManifestVersion = N` means the client is running the deployment as manifest N defined
   it, so the version it reports MUST match the revision it actually holds. A client still on an
   older revision reports that older version, and the WFM sees it has not caught up.
+- Fetching a newer manifest does not by itself advance `observedManifestVersion` for a
+  deployment. Until the client begins applying that deployment's newer revision, it keeps
+  reporting the version of the revision it still holds and is applying or running.
 - The reported version need not be the one at which the deployment last changed. A client that
   has lost its local state and resynchronized reports the version it resynchronized against,
   which can be later. The WFM treats any version at or above its own target for the deployment
@@ -146,6 +149,7 @@ to `targetVersion`, together with `status.state`:
 | at or above | `installing` / `pending` | WFM client has taken up the current desired state and is applying it |
 | at or above | `failed` | WFM client took up the current desired state and failed applying it (the `state` on each component localizes it) |
 | below | `installed` | WFM client is still running a previous desired state; it has not yet picked up the change |
+| below | `installing` / `pending` | WFM client is still applying a previous revision and has not yet started on the current one |
 | below | `failed` | WFM client failed on a previous desired state while a newer one is published |
 
 Because manifest versions never repeat, this comparison stays correct across reverts. After
@@ -154,6 +158,13 @@ Because manifest versions never repeat, this comparison stays correct across rev
 holds; it reports `D2`'s version, which is below `targetVersion`, so the WFM reads it as stale
 rather than converged, even though the current content matches `D1` again. Once the client takes
 up the re-published `D1`, it reports the third version and the WFM reads it as current.
+
+The same rule handles newer content that arrives while an earlier apply is still running.
+Suppose manifest `N` sets a deployment's revision, the client starts applying it, and manifest
+`N+1` changes that deployment again before the first apply finishes. Having fetched `N+1` does
+not advance the reported version: the client still holds and is applying `N`'s revision, so it
+reports `N` with `installing` until it actually begins applying `N+1`'s revision, at which point
+it reports `N+1`. The WFM therefore always knows which revision an `installing` refers to.
 
 The WFM client is required only to report the version it took up and the outcome. Deriving these
 states is the WFM's responsibility.
--- a/proposals/sup_desired_state_read_receipts.md
+++ b/proposals/sup_desired_state_read_receipts.md
@@ -205,6 +205,15 @@ states is the WFM's responsibility.
   from the manifest fetch. Adding a separate generation number to the `ApplicationDeployment`
   would be a change on both the desired state and status sides to convey information that is
   already on the wire. That adds surface area for no gain over the existing version.
+- **Echo the deployment's `digest` in the status report.** The `digest` names the exact content
+  the client applied, so for a single deployment it can answer "is the client on the current
+  target?" by equality, and it needs no rule about when the reported value advances. We report
+  `observedManifestVersion` instead because it is monotonic and a `digest` is not. Ordering is
+  what the WFM interpretation relies on ("at or above target"): it lets the WFM place a report on
+  its publication timeline, tell how far behind a client is, and track rollout progress, none of
+  which equality against the current `digest` gives. Content can also repeat (a revert, or a
+  re-publish of identical content), so a `digest` does not uniquely place a report in time,
+  whereas `manifestVersion` never repeats.
 
 ## Rejection reason
 
--- a/proposals/sup_desired_state_read_receipts.md
+++ b/proposals/sup_desired_state_read_receipts.md
@@ -25,8 +25,8 @@ back which revision a client actually took up. This SUP adds one field to the
 `DeploymentStatusManifest` that the WFM client sends the Workload Fleet Manager (WFM), so that
 each status report says which desired state it is reporting on:
 
-- `observedManifestVersion`: the `manifestVersion` of the most recent state manifest the WFM
-  client has reconciled this deployment against.
+- `adoptedManifestVersion`: the `manifestVersion` of the most recent state manifest the WFM
+  client has adopted for this deployment.
 
 It lets the WFM tell, for each deployment, whether the WFM client has picked up the current
 desired state. Manifest versions increase monotonically and never repeat, so the WFM can place
@@ -61,7 +61,7 @@ each deployment.
 
 The value is already available to the WFM client: it holds the `manifestVersion` from the
 manifest it fetched. This SUP asks the client to record, next to each deployment it reconciles,
-the version of the manifest it reconciled against, and to report it. No new computation and no
+the version of the manifest it adopted, and to report it. No new computation and no
 extra fetch.
 
 ## Requirements alignment acknowledgement
@@ -75,7 +75,7 @@ change keeps the same `deploymentId` and yields a new digest.
 
 What this SUP changes in the specification:
 
-- `DeploymentStatusManifest` gains one field, `observedManifestVersion`.
+- `DeploymentStatusManifest` gains one field, `adoptedManifestVersion`.
 - It reuses a value already defined: the `manifestVersion` on the State Manifest.
 
 Out of scope, including the parts of #156 left to other SUP(s):
@@ -103,19 +103,19 @@ separate SUP.
 
 ### New field on `DeploymentStatusManifest`
 
-Add one required field, `observedManifestVersion`. It is required, so this is a breaking change
+Add one required field, `adoptedManifestVersion`. It is required, so this is a breaking change
 to the status report.
 
 | Field | Type | Required? | Description |
 |-------|------|-----------|-------------|
-| observedManifestVersion | number | Y | The `manifestVersion` (see [State Manifest](https://docs.margo.org/specification/margo-management-interface/desired-state#endpoints---state-manifest)) of the most recent state manifest the WFM client has reconciled this deployment against: the latest manifest whose revision of the deployment the client holds and is applying or running. Manifest versions are monotonic and never repeat across a client's manifest history, so the WFM can order this report against the versions it has published. |
+| adoptedManifestVersion | number | Y | The `manifestVersion` (see [State Manifest](https://docs.margo.org/specification/margo-management-interface/desired-state#endpoints---state-manifest)) of the most recent state manifest the WFM client has adopted for this deployment: the latest manifest whose revision of the deployment the client holds and is applying or running. Manifest versions are monotonic and never repeat across a client's manifest history, so the WFM can order this report against the versions it has published. |
 
 Semantics:
 
-- `observedManifestVersion = N` means the client is running the deployment as manifest N defined
+- `adoptedManifestVersion = N` means the client is running the deployment as manifest N defined
   it, so the version it reports MUST match the revision it actually holds. A client still on an
   older revision reports that older version, and the WFM sees it has not caught up.
-- Fetching a newer manifest does not by itself advance `observedManifestVersion` for a
+- Fetching a newer manifest does not by itself advance `adoptedManifestVersion` for a
   deployment. Until the client begins applying that deployment's newer revision, it keeps
   reporting the version of the revision it still holds and is applying or running.
 - The reported version need not be the one at which the deployment last changed. A client that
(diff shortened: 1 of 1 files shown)


phil-abb:
@matlec - This wasn't published as a draft. It's not required. I just wanted to confirm if it was intentional so I knew what state to put this one in.

matlec:
@phil-abb Thanks for asking! I think the SUP is small enough to skip drafting. No need to rush though.

phil-abb:
@margo/technical-wg 

A new SUP has been proposed to add read receipts to the desired state. The SUP is in final review. Please review the proposal and add any questions/comments to the PR.

proposals/sup_desired_state_read_receipts.md
```diff
@@ -0,0 +1,202 @@
...
+  mandate the reaction.
+- What a client reports for a deployment it never obtained from the desired state, for example
+  a fetch that never succeeded. This SUP covers deployments the client has taken up.
+
+## Technical proposal
+
+This SUP makes one identity rule normative that the specification currently only implies: when
+the WFM changes the content of an existing deployment (its parameters or a component
+`revision`), the `deploymentId` stays the same and the digest changes; a different
+`deploymentId` means a different deployment. The `{deploymentId}/{digest}` fetch endpoint
```
phil-abb, line 97: This doesn't align with my understanding of how the deployment ID (ApplicationDeployment ID) should be used. 

The ApplicationDeployment ID is a unique identifier for each instance of the ApplicationDeployment manifest. Meaning, any time a new one is created for any reason (new deployment, updated parameters, etc.), a new ID is always assigned to it. The same ID is never reused. 

Based on this, the WFM should be able to map exactly what changes are associated with a deployment ID if it is tracking it.

This is why I'm not sure I see the need for adding this  `observedManifestVersion` field to the DeploymentStatus manifests. When the device sends the DeploymentStatus manifests, it is associated with a well-known ApplicationDeployment manifest by the ID, so the WFM knows what it's reporting the status against.

matlec: @phil-abb Let's first agree that the spec today does not decide what constitutes a 'new revision' vs 'same deployment updated' and that this is the discussion we need to have to address that part of [spec issue 156](https://github.com/margo/specification/issues/156).

An `ApplicationDeployment` today carries two distinct values:
- the `deploymentId`, the UUID the WFM assigns to the deployment
- the `digest`, the content hash of the YAML

These are two things the spec keeps distinct: the identity of a deployment (`deploymentId`) and its content at a point in time (`digest`). The endpoint that serves a single `ApplicationDeployment` is already addressable by `digest` alone, which is what the spec relies on to keep fetches immutable. So `digest` on its own is enough to fetch content, and the `deploymentId` in `/deployments/{deploymentId}/{digest}` must be there to name something more durable than a single revision. That only holds if one `deploymentId` points at different digests over its life. A new `deploymentId` on every change would map to exactly one digest, so the API today already leans toward a stable `deploymentId`.

If we settle on the interpretation you describe (a fresh `deploymentId` on every parameter or version change), we run into a problem on the WFM client side. If a change retires an old `deploymentId` and introduces a new one, the client sees the old one as "no longer present" and removes it, and installs the new one. There is no way to let the client know about the intention to *update* a deployment. For the Helm path, for example, that means a client can only ever do `helm uninstall` and `helm install`, never `helm upgrade`. This comes with a lot of consequences, e.g.:
- if the new deployment fails and the old deployment is removed, this means an outage for the app
- stateful workloads might lose their data / volumes or fail to install because the volume for the new instance is still bound to the old instance

I think having an option for rolling upgrades and retaining volumes whenever a change of a parameter or an upgrade of an app is needed is a useful thing to support.

So my proposal is: changing the content / `ApplicationDeployment` YAML of a deployment (parameters, versions, ...) keeps the `deploymentId` stable and produces a new digest, the client reconciles it as an in-place update of the same workload. Under this model, `helm upgrade`, rollbacks, and volume retention are easily expressible. And under this model, we need a read receipt such as the one proposed in the SUP to let the client confirm which revision of a deployment it has reconciled (`deploymentId` + `observedManifestVersion`).

phil-abb: >Let's first agree that the spec today does not decide what constitutes a 'new revision' vs 'same deployment updated' and that this is the discussion we need to have to address that part of https://github.com/margo/specification/issues/156.

Yeah, I agree the spec is not clear in this area. I'm interested in how other people have interpreted the spec up to this point.

>If we settle on the interpretation you describe (a fresh deploymentId on every parameter or version change), we run into a problem on the WFM client side. If a change retires an old deploymentId and introduces a new one, the client sees the old one as "no longer present" and removes it, and installs the new one.

The ApplicationDeployment has the `applicationId` which could be used to identify the application that needs to be updated.

matlec: > The ApplicationDeployment has the applicationId which could be used to identify the application that needs to be updated.

The same app could be deployed multiple times to the same device (e.g. to different namespaces or with different parameters). So the client can't use `applicationId` to know which running instance a change targets.

> I'm interested in how other people have interpreted the spec up to this point.

I agree, let's hear how other people interpret the spec and what's their perspective.

phil-abb: >The same app could be deployed multiple times to the same device (e.g. to different namespaces or with different parameters). So the client can't use applicationId to know which running instance a change targets.

True. I don't know that we've talked about it much, if any, but I can see the need to deploy multiple instances of the same application.

nseba: I think this is a welcome proposal. I agree with @matlec on this. The `deploymentId` should be stable once the WFM creates a new deployment, while the digest can change as result of parameter value changes, internal versions etc. At the end of the day it's the same app being deployed but with changes in the way it is deployed. 

This is inline with the `upgrade` vs `install`/`uninstall` workflows as @matlec pointed out, so the behavior is clear. Changes to an existing deployment will always result in an `upgrade` action, while creating/removing a deployment will result in `install`/`uninstall` actions. This makes it predictable and easy to understand.

On the other hand, changes in the actual YAML content can be caused by many reasons and the digest should always change, making each 'snapshot' of the YAML uniquely identifiable, cacheable and most important, immutable. 

If on the other hand the application deployment changes every time, then there is no way to `upgrade` vs `install`/`uninstall`. And as @matlec mentioned an `install`/`uninstall` process might have a completely different end result compared to an update, especially when it comes to mapped volumes, persisted data etc. + can cause outage in the process. Especially for helm workloads where an update might target a specific resource (e.g. an `Ingress` or `ConfigMap` or `Secret`) while a full uninstall/reinstall would cause the entire helm chart contents to be destroyed and recreated.

As for the read receipts, this would clearly allow the WFM to get a clear status of the actual and desired state, which at the moment is not possible.

nilanjan-samajdar: > So my proposal is: changing the content / `ApplicationDeployment` YAML of a deployment (parameters, versions, ...) keeps the `deploymentId` stable and produces a new digest, the client reconciles it as an in-place update of the same workload. Under this model, `helm upgrade`, rollbacks, and volume retention are easily expressible. And under this model, we need a read receipt such as the one proposed in the SUP to let the client confirm which revision of a deployment it has reconciled (`deploymentId` + `observedManifestVersion`).

I support this definition for defining what constitutes an "Update". This was a topic I was seeking guidance on from the PM Group, but glad to see it being developed here.

phil-abb: >As for the read receipts, this would clearly allow the WFM to get a clear status of the actual and desired state, which at the moment is not possible.

The way things are currently designed, the ApplicationDeployment YAMLs and DeploymentStatus reports are separate documents for each app deployment. The state manifests combine all app deployments under a single document with a single manifest version. If we want to be able to associate the ApplicationDeployment and DeploymentStatus with a specific manifest version, does it make more sense to have a single application deployment document representing all the application deployments for the state manifest version, and a single deployment status document covering the statuses for all app deployments for that specific state manifest?

matlec: > If we want to be able to associate the ApplicationDeployment and DeploymentStatus with a specific manifest version, does it make more sense to have a single application deployment document representing all the application deployments for the state manifest version, and a single deployment status document covering the statuses for all app deployments for that specific state manifest?

A single `observedManifestVersion` on a combined document works only if `observedManifestVersion` means the manifest the client has *fetched*. The value we actually need (and what I meant in the SUP) is which revision each deployment is currently applying or running.

Take a deployment D at digest1 in manifest N. The client starts the `helm upgrade` toward digest1. Before it finishes, manifest N+1 changes D to digest2 and the client fetches it. If the status just carried "I've fetched N+1", it would report N+1 with D `installing`, and the WFM would read that as D installing digest2 when it is still applying digest1. And another deployment in the same manifest may already be on N+1, so a single document-level value would have to be N and N+1 at the same time.

So `observedManifestVersion` has to report the version whose revision each deployment actually holds and is applying, which for D stays at N until it really starts on digest2. That advances independently for each deployment and lags the fetched manifest during any rollout, so it has to be reported for each deployment rather than once for the document, even if the WFM accepts a single status document. I'll extend the SUP to make this explicit, good point!

On your question: I'd keep the documents separate (and I think changing the structure would be out of scope for this SUP):
- Splitting the desired state into content-addressed deployment YAMLs lets a client fetch only what changed, which is helpful for metered or bandwidth-limited links. And we already have the bundle option for fetching all deployments at once
- Deployments converge at different times and report as they go, so a single combined status document makes the client re-send the full set on every change

phil-abb: > Splitting the desired state into content-addressed deployment YAMLs lets a client fetch only what changed, which is helpful for metered or bandwidth-limited links. And we already have the bundle option for fetching all deployments at once

Yeah, this is not part of this scope, but I'll probably come back to this idea at some point. It depends on how some of these other in-progress SUPs turn out.

>Deployments converge at different times and report as they go, so a single combined status document makes the client re-send the full set on every change

Doesn't this run the risk of putting things in an unknown state?

For example, the device pulls down the latest desired state manifest with manifest version 5. The device determines it needs to install two new applications and update one application. 

The device installs the two applications and independently reports deployment status results for each of these deployment IDs for manifest version 5. At this point, the WFM knows the device has completed the two installs requested for manifest version 5. 

However, for some reason, the WFM never receives the report for the deployment status for the update that was supposed to happen for manifest version 5.  At this point, the WFM is in an uncertain state because it only knows part of what has occurred. 

It seems like sending a single deployment status report for desired state manifest version 5 with the results of every install, update, or removal for that manifest version would be better. 

From a WFM supplier's perspective, which is better? Getting independent reports for each deployment ID, or getting a single report with the results for applying a specific desired state manifest version? If we keep it independent and the WFM doesn't receive all the deployment status reports it expects, is this a concern?

matlec: > Doesn't this run the risk of putting things in an unknown state?
> ...
> However, for some reason, the WFM never receives the report for the deployment status for the update...

If the client resends the full set on every change, the WFM actually holds a complete snapshot each time. But that doesn't remove the risk: if the last snapshot, the one where the update finished, is lost and nothing changes after it, the WFM sits on a stale view until the next change triggers a resend. So this stays a delivery problem.

Either way, the `observedManifestVersion` this SUP adds still has to live in the status report. So independent vs combined is a question of reporting format, and I'd rather settle that in a different SUP than fold it in here. Nothing in this SUP rules out the snapshot.

> From a WFM supplier's perspective, which is better?

From a WFM perspective, I'd favor independent reports. A WFM likely keys deployment state on (client, deploymentId), so independent status reports map directly onto that. It also scales better for fleets with many deployments: a combined status report re-sends the whole set on every change, so the WFM keeps ingesting and re-processing information it already has.

proposals/sup_desired_state_read_receipts.md
```diff
@@ -0,0 +1,222 @@
...
+[@matlec](https://github.com/matlec)
+
+## Summary
+
+Today the WFM publishes a desired state but never hears
+back which revision a client actually took up. This SUP adds one field to the
+`DeploymentStatusManifest` that the WFM client sends the Workload Fleet Manager (WFM), so that
+each status report says which desired state it is reporting on:
+
+- `observedManifestVersion`: the `manifestVersion` of the most recent state manifest the WFM
```
phil-abb: Is "observed" needed in the field name? It doesn't seem to fit because the device is doing more than "observing".  I'm not sure if we need anything other than "manifestVersion" or possibly "desiredStateManifestVersion" if we don't feel "manifestVersion" is clear enough about what manifest is being referred to.

matlec: > Is "observed" needed in the field name?

Good push :) Thinking about it more, "observed" isn't quite right: as discussed [above](https://github.com/margo/specification-enhancements/pull/77#discussion_r3587195097), during an upgrade the client keeps reporting the older version it's still applying even after it might have fetched a newer one.

I'd avoid plain `manifestVersion` though as it conveys little semantics. `desiredStateManifestVersion` doesn't separate what the WFM published from what the client took up, and "desiredState" reads a bit oddly on the status side, which is the actual state rather than the desired one.

What the field really means is the version the client has taken up and is acting on. Maybe something like `adoptedManifestVersion`?

phil-abb: `adoptedManifestVersion` is ok. What about `targetManifestVersion`?

matlec: `targetManifestVersion` reads as the version you're aiming for with the deployment, but this field expresses the version the client actually has taken up *and* is acting on, so it may lag behind the actual "target" version. I'll go with `adoptedManifestVersion` unless you feel strongly about `target`.

phil-abb: No strong feelings. `adoptedManifestVersion` is fine with me.

proposals/sup_desired_state_read_receipts.md
```diff
@@ -0,0 +1,222 @@
...
+Add one required field, `adoptedManifestVersion`. It is required, so this is a breaking change
+to the status report.
+
+| Field | Type | Required? | Description |
+|-------|------|-----------|-------------|
+| adoptedManifestVersion | number | Y | The `manifestVersion` (see [State Manifest](https://docs.margo.org/specification/margo-management-interface/desired-state#endpoints---state-manifest)) of the most recent state manifest the WFM client has adopted for this deployment: the latest manifest whose revision of the deployment the client holds and is applying or running. Manifest versions are monotonic and never repeat across a client's manifest history, so the WFM can order this report against the versions it has published. |
+
+Semantics:
+
+- `adoptedManifestVersion = N` means the client is running the deployment as manifest N defined
```
nilanjan-samajdar: Corner-case : What should be the value of `adoptedManifestVersion` reported, when the very first instance of the application failed to come up ?
In this case there is no previous version to report, as this is  the first instance for the application being installed.
Should we make this type a nom-zero number, with zero being used for this special case ?

nilanjan-samajdar (changes requested):

proposals/sup_desired_state_read_receipts.md
matlec: Even if the first ever version of a deployment fails, the client still has a version to report: the `manifestVersion` it took up that first contained that deployment. On a first install the client fetched that revision and tried to apply it, so it reports that manifest version with `status.state: failed` (the "at or above target + failed" row). So I don't think we need a special "zero sentinel". Every deployment the client has taken up has a real version.

Does that address your concern?

matlec: @nilanjan-samajdar did you have the chance to review my feedback to your comment? Please let me know whether this already addresses your concern or if you had something different in mind. Thanks!

nilanjan-samajdar: I would recommend adding this interpretation for the first ever version to the documentation of this attribute.

matlec: @nilanjan-samajdar Agreed! I added a clarification and also defined the meaning of "adoption" more clearly. Thanks for your review!
If you're happy with the changes, please approve.

nilanjan-samajdar (approved):

phil-abb:
This SUP has been approved, so I'm force-merging this PR: https://github.com/margo/specification-enhancements/issues/82

PR #84, opened 2026-07-28
[DRAFT] MIAF Credential Provisioning and Acquisition

matlec:
This SUP standardizes how a Margo principal acquires its X.509-SVID and trust material, and what it must accept as provisioning input, independently of any enrollment protocol. It defines three acquisition modes (installed, enrolled, delivered) with a conformance floor for each, an input contract with one interchange format for each artifact class, the rules for the initial bootstrap anchor, and MIS independence: no principal may require a specific MIS implementation.

It builds on the MIAF framework and the WFM Identity Profile and does not change them. The forthcoming (not yet published) enrollment SUP builds on top of it: its protocol automates the enrolled mode.


phil-abb:
[@margo/technical-wg](https://github.com/orgs/margo/teams/technical-wg) A new SUP has been started by [@matlec](https://github.com/matlec) for MIAF credential provisioning and acquisition. If anyone is interested in helping work on fleshing out the details of this SUP, please contact [@matlec](https://github.com/matlec)

proposals/miaf-credential-provisioning-and-acquisition.md
```diff
@@ -0,0 +1,299 @@
...
+
+<sup>*</sup> The artifact is not a provisioning input in that mode. The delivered principal receives it from the SPIFFE Workload Endpoint instead.
+
+**Both principals:**
+
+| Input | Notes |
+| :---- | :---- |
+| X.509-SVID + intermediate chain | the principal's own SVID with its chain (leaf + intermediates), per the X.509-SVID profile; installed mode (with key) or enrolled mode (the certificate chain returned by the CA) |
+| Private key (only when operator-supplied) | installed mode only<sup>1</sup> |
+| Initial trust anchors | one or more certificates; PKI-anchored bootstrap: the principal validates the MIS server chain to these anchors, with RFC 6125 name checks<sup>2</sup> |
```
FHatCSW: RFC 9525 obsoletes RFC 6125

proposals/miaf-credential-provisioning-and-acquisition.md
```diff
@@ -0,0 +1,299 @@
...
+
+Normative rules:
+
+- **Mode declaration (every principal).** A principal MUST implement at least one acquisition mode and MUST declare (in a product's conformance documentation) which of the three modes it implements. It MAY implement several. The operator chooses among them for each deployment.
+- **Bundle floor (installed and enrolled modes).** Every principal in the installed or enrolled mode MUST accept a statically supplied Trust Bundle ([§4](#4-provisioning-input-contract) format), and SHOULD additionally support `trustBundleUri` refresh. To refresh, a principal retrieves the Trust Bundle from its `trustBundleUri` repeatedly while it operates, per the specification's rules for Trust Bundle retrieval. Each retrieval replaces the bundle the principal holds. A delivered principal acquires its Trust Bundle over the Workload API instead. With a static-only bundle, every trust-anchor rotation requires re-provisioning every principal. With `trustBundleUri` refresh, rotation propagates on its own (see the MIAF trust-anchor rotation playbook).
```
FHatCSW: Please require validation before activation + retention of the previous valid bundle for controlled recovery.

proposals/miaf-credential-provisioning-and-acquisition.md
```diff
@@ -0,0 +1,299 @@
...
+- **Exactly three acquisition modes.** A principal's SVID acquisition is one of three named modes, distinguished by who generates and first holds the private key: **installed** (the operator generates the key and the SVID and places both into the principal), **enrolled** (the principal generates its own key and is issued a certificate for it), and **delivered** (SVID, key, and Trust Bundle arrive over a local [SPIFFE Workload API](https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Workload_API.md)). The enrolled round-trip is a PKCS#10 certificate request answered by a certificate chain. It runs by hand today, and later through the forthcoming enrollment protocol, which a sibling SUP on the MIAF roadmap defines ([§9](#9-roadmap-and-deferred-items-informative)). A principal declares which of the three modes it implements, and each mode carries a small conformance floor, a minimum capability every implementation of that mode must have. Naming the modes puts manual provisioning, local SVID delivery, and future automated enrollment on equal footing: each is a conformant way to acquire an SVID. The floors keep provisioning portable across vendors ([§3](#3-acquisition-modes)).
```
FHatCSW: The three modes (manual, localSVID, automated) may be equally conformant from an interoperability perspective, but they are not equivalent from a security-assurance perspective (private key exposure).

proposals/miaf-credential-provisioning-and-acquisition.md
```diff
@@ -0,0 +1,299 @@
...
+- **CSR semantics (enrolled mode).** The round-trip's PKCS#10 CSR carries the principal's public key and proof of possession. Any subject or SAN content it carries is advisory. The issuer is authoritative for the SPIFFE ID. The operator mints the SVID for the chosen SPIFFE ID and overrides any name content in the CSR (per the MIAF operator provisioning playbook). A principal is therefore issued its identity rather than requesting it, and never needs to know it in advance (see the identifiers note in [§4](#4-provisioning-input-contract)). A renewal CSR MAY carry the principal's current SPIFFE ID. That content is advisory too.
+- **Chain ingestion (enrolled mode).** On ingesting the returned chain, the principal MUST verify that the leaf certifies its own key pair, and that the leaf carries a well-formed SPIFFE ID whose shape matches an identity profile the principal implements. It SHOULD validate the chain against its current Trust Bundle when it holds one. Failures of any of these checks are surfaced as the validation rule in [§4](#4-provisioning-input-contract) requires.
```
FHatCSW: I would turn SHOULD into MUST with a complete RFC 5280 certification-path validation

proposals/miaf-credential-provisioning-and-acquisition.md
```diff
@@ -0,0 +1,299 @@
...
+
+<sup>3</sup> **Discovery cross-check.** Where discovery is used, a principal that already holds an SVID SHOULD verify the document's `trustDomain` against its own SPIFFE ID's trust domain, per the specification's discovery rules. This catches a principal pointed at the wrong Trust Domain's document on an origin serving several.
+
+<sup>4</sup> **Accepted-client policy.** This row states only that a WFM MUST be configurable with the set of client identities it authorizes. The WFM Identity Profile's authorization section leaves the expression of policy implementation-specific. This row deliberately strengthens that: being configurable with an explicit list becomes a conformance requirement. Accepting a whole namespace by wildcard remains permitted alongside it. Revocation by removal still needs individual entries, as the profile's lifecycle section notes. How an operator *manages* that set at runtime is a management-interface concern between an operator and its WFM, which Margo does not specify, and is out of scope here. No interoperable import payload is defined, and the entries are supplied as a provisioning input through each product's documented acceptance surface.
```
FHatCSW: Removing a SPIFFE ID from a WFM accepted-client policy is application-level authorization withdrawal? It is not X.509 certificate revocation...

proposals/miaf-credential-provisioning-and-acquisition.md
```diff
@@ -0,0 +1,299 @@
...
+- **Public-PKI case.** Where the MIS trust endpoints are already reachable over publicly-trusted PKI with a resolvable DNS name, the initial bootstrap anchor is the principal's built-in trust store, so the operator provisions no anchor at all, only the endpoint URL. This SUP does not recommend exposing an MIS on public PKI where it would not otherwise be.
+- **Private or pinned mode.** Where the MIS is private or air-gapped, the operator MUST provision an anchor set or pins out of band, per the specification's rules for initial trust bootstrap. Those rules fix the pin interchange form (formats table, [§4](#4-provisioning-input-contract)), so pins produced by one operator's toolchain are ingestible by any conformant principal. Operators SHOULD pin the SubjectPublicKeyInfo of a CA rather than of a leaf certificate, since leaf pins break on routine MIS certificate renewal.
```
FHatCSW: Would this allow every endpoint to authorize issued/certified by that CA?

Maybe add additional endpoint-identity validation within the pinned CA scope

proposals/miaf-credential-provisioning-and-acquisition.md
```diff
@@ -0,0 +1,299 @@
...
+
+**The out-of-band channel as a role (informative).** The precondition above names an out-of-band channel without fixing a procedure for it. In this SUP the operator fills it by hand, installing the anchor (and, under the enrollment profile, the enrollment authorization) through its existing provisioning path. A later device identity profile can fill the same role with an automated mechanism rooted in the supply chain, delivering the same two inputs against a factory identity: FIDO Device Onboard, BRSKI ([RFC 8995](https://datatracker.ietf.org/doc/html/rfc8995)), or a vendor-specific bootstrap protocol ([AOKI](https://trustpoint.readthedocs.io/en/latest/features/aoki/index.html), for example). Each is a *producer* of the bootstrap inputs defined here, and the acquisition and enrollment layers are the *consumer*. Admitting one requires no change to this SUP, provided the inputs stay defined as roles rather than as one specific delivery.
```
FHatCSW: IEEE 802.1AR-based identities should be the preferred interoperable model.

factory identity or IDevID → ownership authorization → LDevID or enrollment credential → workload SVID

proposals/miaf-credential-provisioning-and-acquisition.md
```diff
@@ -0,0 +1,299 @@
...
+
+- **Identifier reuse.** A WFM Client is revoked by removing its SPIFFE ID from the WFM's accepted-client policy (see the MIAF operator revocation playbook). The removal acts on the ID, so it cuts off every credential that carries it. When a replacement principal reuses the predecessor's `wfm-client-id`, the two credentials can no longer be revoked separately. Reusing a `wfm-client-id` is therefore safe only when the predecessor credential can never be presented again (its key is destroyed or unrecoverable). Otherwise the operator assigns a fresh `wfm-client-id`: removing the predecessor's entry then revokes it without affecting the replacement. The re-issuance workflow of the MIAF operator provisioning playbook leaves the reuse choice to operator policy. The guidance above gives the operator a default.
```
FHatCSW: This action withdraws application authorization but does not revoke the X.509 certificate.

--> Key destruction and policy removal do not eliminate the need for PKI revocation.

proposals/miaf-credential-provisioning-and-acquisition.md
```diff
@@ -0,0 +1,299 @@
...
+
+### 9. Roadmap and Deferred Items (Informative)
+
+This section gives a consolidated view of the MIAF roadmap.
+
+- **Automated enrollment, renewal, and revocation (forthcoming enrollment profile).** The enrollment profile automates the *enrolled* mode of [§3](#3-acquisition-modes): its protocol runs the same PKCS#10 round-trip over the wire, for first issuance and for renewal, and its enrollment authorization is the second static bootstrap input the [§5](#5-the-initial-bootstrap-anchor) precondition already covers. It also covers revocation, including revocation status that verifiers can check. The installed mode and its operator-supplied keys remain available outside the profile as the manual fallback.
```
FHatCSW, line 275: The future enrollment profile should select and profile established PKI protocols rather than define only a proprietary CSR transport.

CMP according to RFC 9810 is a strong fit for a complete industrial certificate lifecycle

proposals/miaf-credential-provisioning-and-acquisition.md
```diff
@@ -0,0 +1,299 @@
...
+
+**WFM only:**
+
+| Input | Notes |
+| :---- | :---- |
+| Accepted-client policy entries | the WFM Client identities this WFM authorizes, as a list of identifiers (`wfm-client-id` or full SPIFFE ID); list serialization is product-specific<sup>4</sup> |
+
+<sup>1</sup> **Key handling.** In the installed mode the operator-supplied key interchanges as unencrypted PKCS#8 PEM. Encryption of the key file is deliberately outside the contract, since confidentiality on the key path is the provisioning channel's job (the playbook already requires it) and passphrase handling would otherwise become a provisioning input of its own. A principal SHOULD prefer the enrolled mode over the installed mode where its hardware allows: the key pair is generated on the principal, bound to hardware (TPM, secure element, HSM) where available, and never leaves it. Both modes are conformant, consistent with the operator provisioning playbook, which allows centrally generated keys as a fallback. Key generation and CSR signatures follow the MIAF cryptographic requirements, including the requirement to generate keys with a cryptographically secure random number generator.
```
FHatCSW: Maybe explicitly state that if a secure storage is available the keys should be non exportable.

Maybe the requirement should be stronger for principals CAPABLE of secure local key generation

FHatCSW: I could not find the MIAF cryptographic requirements. 

They should at least define:

- permitted signature algorithms
- minimum RSA or elliptic-curve key sizes
- approved curves
- permitted hash functions
- TLS cipher suites and key-exchange groups

matlec: @FHatCSW thanks for taking the time to review! I will address your other comments later, but just wanted to point you to [this PR](https://github.com/margo/specification/pull/194) that integrates MIAF with the Margo spec. This PR pins the cryptographic requirements you're asking for.

FHatCSW: @matlec Thanks for the hint. I will try to review also this one

proposals/miaf-credential-provisioning-and-acquisition.md
matlec: thanks, good catch

proposals/miaf-credential-provisioning-and-acquisition.md
matlec: Agreed. Both rules are [in the spec now](https://github.com/margo/specification/pull/194/commits/35128226c3975188c10620b64ee2c3d9bd878f18): a client keeps validating against its current bundle until it accepts a retrieved one, and never validates against a bundle it rejects.

proposals/miaf-credential-provisioning-and-acquisition.md
matlec: Agreed. This is already stated in the SUP elsewhere, but the summary is not clear enough about this. I added a line to the summary.

proposals/miaf-credential-provisioning-and-acquisition.md
matlec: Agreed, removing the SPIFFE ID entry withdraws authorization at the application layer (the certificate stays cryptographically valid). I'll fix the wording and use "authorization withdrawal" instead of "revocation" to avoid any confusion (and will also fix this in the spec PR) :)

proposals/miaf-credential-provisioning-and-acquisition.md
matlec: Same fix as the [thread above](https://github.com/margo/specification-enhancements/pull/84#discussion_r3680826028). On the need for PKI revocation: agreed on the need, however, online revocation is deferred to the enrollment SUP.

proposals/miaf-credential-provisioning-and-acquisition.md
matlec: Yes, that's the consequence. I've addressed your comment [with this commit](https://github.com/margo/specification-enhancements/pull/84/commits/25ef7744541e9472156bb7884a92807b95199974).

proposals/miaf-credential-provisioning-and-acquisition.md
matlec: Good point, I have added a remark on non-exportability.
On making this requirement normative (SHOULD / MUST): Right now, MIAF takes the following stance:
> An identity profile MAY further constrain validity periods, key-protection rules, or path conventions for its own principal classes.

I would expect the future device identity profile to make stronger recommendations or obligations on key protection. I wouldn't make this a framework-level requirement or a requirement for WFM server or client keys. A key protection requirement is also unverifiable from outside the device until attestation exists, which might arrive with the same profile.

proposals/miaf-credential-provisioning-and-acquisition.md
matlec: The forthcoming enrollment SUP will profile an established protocol (and will not define a Margo-specific protocol). The candidates I have in mind are EST, ACME and CMP. I'm currently leaning towards EST, but this is still an open topic and I expect this to be a tougher decision with lots of opinions :) That's why I deliberately deferred the protocol decision to a different SUP. Automated enrollment will be needed for fleet scale deployments, but in my view, the contract defined in this SUP would be sufficient for a GA release and to start with the first pilots.

I agree that CMP is a strong candidate, especially its PKI lifecycle coverage. My biggest concern around CMP specifically is its complexity combined with the lack of maintained client libraries in ecosystems like Go and Rust. Happy to hear your thoughts and experiences (also offline) in this area.

proposals/miaf-credential-provisioning-and-acquisition.md
matlec: I've added a reference to IEEE 802.1AR which I kind of implied with "factory identity". But it's better to make that more explicit, thanks! However, I don't want to introduce normative language regarding 802.1AR in this SUP. Bootstrap mechanisms and factory identities should be covered by the future device identity profile.

chrisgclayton:
Overall, I agree with the concept and direction of the proposal. I have a few questions and suggestions that I believe could improve clarity and interoperability.

1.) If we are defining a WFM and WFM Client model, are we implicitly recommending that Margo remain agnostic to the previous DFM concept and instead focus only on the artifacts that a DFM must produce? If that is the intent, it may be helpful to state it explicitly.

2.) I would recommend considering an additional onboarding artifact that identifies the WFM endpoint (or rendezvous service) the device should use. I do not believe the specification needs to mandate how this information is obtained, only that the device can determine where it should connect.

3.) Should the specification recommend that devices store credentials using the most secure storage mechanism available? Additionally, should devices publish information about how credentials are protected (for example, hardware-backed storage, encrypted software storage, etc.) so that WFMs can enforce onboarding policies based on storage guarantees?

4.) For credential delivery methods that include private key material, it may be worth adding a non-normative recommendation that modern secure key exchange mechanisms be used. This would not need to be a requirement but could provide useful security guidance.

5.) My understanding is that the trust domain cross-check is not applied between the Device ID and the WFM Client ID. Is that correct? If so, could this create issues in scenarios where the Device ID exists outside the WFM trust domain?

6.) If DFM-based onboarding is a desired interoperability scenario, I would suggest defining a minimum required DFM standard (regardless of which one is selected) so that all compliant devices can support at least one common onboarding approach.

7.) The proposal references JWK Sets. If X.509 SVIDs are being used as the primary identity mechanism, could you clarify the role of the JWK Set and whether both are required?

8.) I would recommend reconsidering the use of client IDs within URLs. Since the client identity is already established by the SVID, it seems preferable to derive identity from the presented credentials rather than from URL path parameters.

9.) There is a comment describing MIS deployments as "private or air-gapped." I understand the "private" case, but could you clarify what "air-gapped" means in this context? Does it simply mean that the device communicates only with the WFM, or is a stricter isolation model intended?

Thanks for putting this proposal together. I think the overall direction is solid, and these comments are primarily intended to clarify expectations and improve interoperability across implementations.

matlec:
@chrisgclayton Thanks a lot for your feedback, much appreciated!

1.) I'm not sure to what exactly you're referring to when you say the "previous DFM concept". The DFM is a [*future* component of the specification](https://docs.margo.org/personas-and-definitions/technical-lexicon#device-fleet-manager), and Margo is not agnostic to it as a component: MIAF [names it explicitly](https://github.com/margo/specification/blob/1ec6ee11d74963c2aa01ff5342261121ea5d0e8d/system-design/specification/identity/identity-framework.md#scope-and-applicability) as a Margo component that may hold an identity.

That said, at the layer of this SUP your reading is right: the SUP sees a DFM only through the artifacts it produces. The out-of-band channel in §5 is defined as a role, with mechanisms like FDO as producers and the acquisition modes as the consumer.

The reason the SUP only enumerates provisioning inputs for WFMs and WFM Clients is that we only have an identity profile for these two classes today, and no DFM interfaces defined in the first place :) The general rules (§3, §5, §6) already apply to any future principal class. A DFM identity profile would only add its own input tables to §4. I will add a statement to the SUP to make this clear.

2.) The WFM endpoint URL is already a required provisioning input for the WFM Client (see §4). How it is obtained is deliberately not mandated, delivery is the job of the operator's provisioning channel (e.g. FDO). But maybe I was missing your point: were you asking for a concrete artifact such as a JSON document?

3.) On secure storage: the note in §4 already recommends this:

> Where the principal has hardware-backed key storage (TPM, secure element, HSM), the key can be generated inside that storage and marked non-exportable, so it never exists outside the hardware.

This was sharpened after [@FHatCSW's review](https://github.com/margo/specification-enhancements/pull/84#discussion_r3680925719). I've kept it a recommendation at this level: MIAF lets each identity profile constrain key-protection rules for its own principal classes, and I'd expect the future device identity profile to define stronger obligations.

On devices publishing how their credentials are protected: I agree with the goal, @stormc brought up a similar idea in an earlier MIAF-related PR. However, without attestation such a statement is self-reported, so a WFM enforcing onboarding policy on it gains little real assurance. In my view, attestation is device identity profile territory.

4.) Agreed, I've added a recommendation in the SUP.

5.) Today, we don't have a MIAF device identity profile, so device identities are not governed by MIAF today. The only identities MIAF defines today are the WFM and the WFM Client, and those two *are* cross-checked (same Trust Domain, client named under its WFM's namespace).

On the remaining part, to be on the same page about my vision for MIAF: a device would receive an identity governed by MIAF in the same Trust Domain as the WFM - like every MIAF principal, since MIAF scopes recognition to a single Trust Domain (at least today, federation across domains is deferred). Specifying that identity, and the cross-checks between it and the WFM Client identity, is the job of the device identity profile.

What can come from *outside* the Trust Domain is the factory credential (an IDevID, for example). That is not the device's identity in Margo terms, it's the input the device uses to obtain one, the *enrollment authorization* of §5. What the device identity profile has to define is what the issuer verifies at that point.

6.) I agree on this goal. I hate referring to the device identity profile again, but: selecting and mandating a bootstrap mechanism (FDO, BRSKI, ... whatever it will be) is the core decision of the device identity profile, so I'd keep it out of this SUP. The contract in this deliberately treats bootstrap mechanisms as producers of the same inputs, so the profile can add such a floor later without changing anything here.

7.) The JWK set is the container format the [SPIFFE bundle format](https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Trust_Domain_and_Bundle.md#41-jwk-set) defines for trust material. X.509-SVID is the only accepted credential in Margo (MIAF uses no JWT-SVIDs). I've added a clarification to the SUP.

8.) I've already removed the `clientId` from the entire API surface in the [MIAF integration PR](https://github.com/margo/specification/pull/194) and re-worked the [identity and authentication](https://github.com/margo/specification/blob/1ec6ee11d74963c2aa01ff5342261121ea5d0e8d/system-design/specification/margo-management-interface/api-requirements-and-security.md#identity-and-authentication) section for the Margo Management API.

9.) Thanks, good catch. That wording was absolutely ambiguous (and it also conflated reachability with trust). What the section means is: if the MIS server certificates validate to the principal's built-in trust store, no bootstrap anchor needs to be provisioned. Otherwise (a private or enterprise PKI, for example) the operator provisions an anchor set or pins out of band. I changed the wording in the SUP.

I'll be on vacation for the next 2 weeks, please expect some delay for follow-up discussions.

phil-abb:
In general, I think this makes sense as an interim step to help progress things further and make it easier for integrations during the upcoming plugfest. 

There are a few concerns I have, but I don't think they need to be addressed in this SUP:

#### Nothing is prescribed

This SUP doesn't propose a minimum implementation each Margo-conformant device and customer environment must support. It provides three options and leaves it up to both the device supplier and customer/operator to pick and choose what they want. If we don't eventually agree on at least one minimum way, this will lead to fragmentation.

#### SPIFFE integration

I'm not convinced that having the Delivered mode that uses the SPIFFE Workload API and SPIFFE Workload Endpoint makes sense because of the requirements each of these has.

I think the API defined in the SPIFFE Workload API makes sense, but the API definition is only for gRPC right now. There has been some previous discussion about potential challenges with gRPC.

The [transport requirements](https://spiffe.io/docs/latest/spiffe-specs/spiffe_workload_endpoint/#3-transport) for the SPIFFE Workload Endpoint seem to be too restrictive for the type of deployments Margo is targeting. I'm not an expert on this, so maybe I'm misunderstanding the implications.

There are two acceptable options for exposing the workload endpoint:
- Unix Domain Socket (UDS) - Which means the only workloads that can interact with this endpoint are ones running on the same machine
- TCP listen socket, with the qualifying text "TCP transport MUST NOT be used unless the underlying network allows the Workload Endpoint server to strongly authenticate the workload based on source IP address (e.g., over a localhost or link-local network), or other strong network-level assertions (e.g., via an SDN policy)." - The implications of these seem to be too restrictive for how environments are setup. 
- Also, in the [Accessibility](https://spiffe.io/docs/latest/spiffe-specs/spiffe_workload_endpoint/#2-accessibility) section, it indicates "As a result, the SPIFFE Workload Endpoint SHOULD be exposed through a local endpoint, and implementers SHOULD NOT expose the same endpoint instance to more than one host." 

Based on how I'm understanding the SPIFFE workload endpoint requirements, it seems like its primary use case is for workloads on the same host and not workloads spread out across the network.

I would be interested in hearing if this was something anyone was considering, or interested in trying out. I don't mind keeping it as part of the SUP to see if it's something people want to try, but I think we'll eventually need to determine if it makes sense for Margo's use case.

matlec:
@phil-abb Thanks a lot for your input! I've added a couple of clarifications to the SUP based on your feedback.

On "Nothing is prescribed":
I agree on the risk. However, I'd rather define the minimum baseline / floor in the forthcoming enrollment SUP. The floor proposed for the enrolled mode in this SUP is manual and will not scale. What I'd propose in the enrollment SUP (defining an enrollment profile):
- A WFM and a WFM Client MUST each acquire its SVID by one of two paths: enrollment through the enrollment profile, or this SUP's delivered mode, in which a local SPIFFE Workload Endpoint supplies the SVID.
- A principal that cannot generate a key pair safely MUST implement the installed mode instead, and its conformance documentation MUST state the reason.
- A principal that automates enrollment MUST use the enrollment profile's protocol binding, so at most one automated enrollment protocol exists in Margo.

On the SPIFFE Workload API:
Your reading of the transport rules is correct (per my understanding). Using the delivered mode implies adding a gRPC dependency. Whether this consequence is acceptable for this mode, is open to community debate :)

> Based on how I'm understanding the SPIFFE workload endpoint requirements, it seems like its primary use case is for workloads on the same host and not workloads spread out across the network.

This is correct. The process that is implementing the workload API is expected to run on the same host. However, just to be clear about this: the workload API is not intended as a way to distribute SVIDs across a network. If we take SPIRE as an example for an implementation of the delivered mode, the SPIRE agent would send CSRs to the SPIRE server (over the network) to obtain SVIDs for the (local) workloads. When workloads request their SVID via the Workload API from the SPIRE agent, the SPIRE agent will return the cached SVIDs obtained from the server earlier.

matlec:
@chrisgclayton To your point 2: I have added an explicit requirement to provide the WFM endpoint URL as input. You were right, this was not called out explicitly before.

matlec:
@phil-abb @chrisgclayton With the latest commit, after spending more thought on it, I have removed the *delivered mode* (the SPIFFE Workload API path) from the SUP.

My original intention with this mode was to support the scenario where an edge device (under a future device identity profile) enrolls with the MIS through a Margo-defined enrollment protocol, and where a device-local agent serves a SPIFFE Workload Endpoint to the WFM Client (and other workloads in the future). A different way to implement the *delivered mode* today would be a SPIRE agent serving that endpoint (as described in my comment above). This is what concerns me: a SPIRE agent talks to its SPIRE server over a project-internal protocol, not a standard. If we make such an implementation Margo-conformant now, before an interoperable, automated enrollment alternative exists, we encourage vendors to commit to implementations that do not interoperate at ecosystem level.

For a vendor that builds on SPIRE today, the WFM Client (or WFM) must implement the enrolled mode (with the forthcoming enrollment SUP offering automation for this mode). The SPIRE server can stay in the picture as the CA behind the MIS, the server's API (`MintX509SVID`) can be used to mint an SVID using a CSR. What is out, until the device identity profile, is the SPIRE agent as the component that hands SVIDs to the WFM Client or WFM over the SPIFFE Workload API.

I still think that WFM Clients and workloads eventually should retrieve their SVID and Trust Bundle via the SPIFFE Workload API. But I would defer this mode until the device identity profile and a standard for automated enrollment and renewal are in place.

matlec:
With the latest commit I have rewritten the SUP for readability and improved the alignment with MIAF terminology established in the current spec. I've also renamed "acceptance surface" to "provisioning interface" to align with existing terminology around provisioning channel and provisioning input.

phil-abb:
@margo/technical-wg 

The SUP is in final review. Please review the proposal and add any questions/comments to the PR.

